PR #2, opened 2017-08-03
Add basic integration test suite

joseph-wakeling-sociomantic:
These patches add support for a basic integration test suite that performs multi-class logistic regression on the MNIST dataset using the D MXNet wrapper library.  It does not provide comprehensive coverage (and is not intended to), but should provide a basic sanity check of the library's behaviour.

A script and build target to download the MNIST dataset have been added, and the docker build script has been updated to ensure that the CI docker image will include the dataset already baked in.  This should help to minimize the amount of downloading required (for our sake and MNIST's).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ sudo: required
 services:
     - docker
 
+cache:
+    directories:
+        - $HOME/.cache/mnist/
+
 # Disable automatic submodule fetching to prevent
 # it from being done recursively
 git:
@@ -26,6 +30,8 @@ env:
         - DC=dmd-transitional DIST=xenial F=production
         - DC=dmd-transitional DIST=xenial F=devel
 
-install: beaver install
+install:
+    - beaver install
+    - beaver run script/download-mnist
 
 script: beaver run ci/run.sh
--- a/Build.mak
+++ b/Build.mak
@@ -15,3 +15,7 @@ endif
 .PHONY: download-mnist
 download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
+
+$O/test-mxnet.stamp: override LDFLAGS += -lz
+$O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
+$O/test-mxnet.stamp: download-mnist
--- a/Build.mak
+++ b/Build.mak
@@ -11,3 +11,7 @@ endif
 # use NaiveEngine (non-threaded) MXNet engine since the default (threaded)
 # version dead locks
 %test: export MXNET_ENGINE_TYPE=NaiveEngine
+
+.PHONY: download-mnist
+download-mnist: $C/script/download-mnist
+	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
--- a/docker/build
+++ b/docker/build
@@ -16,4 +16,5 @@ apt-get update
 
 # Install dmxnet dependencies
 apt-get install -y \
-    libmxnet
+    libmxnet \
+    zlib1g-dev
--- a/script/download-mnist
+++ b/script/download-mnist
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+set -eu
+
+download_dir=${1-${MNIST_DATA_DIR-"${HOME}/.cache/mnist"}}
+
+base_url="http://yann.lecun.com/exdb/mnist"
+
+mnist_sha256sums=$(cat <<EOT
+8d422c7b0a1c1c79245a5bcf07fe86e33eeafee792b84584aec276f5a2dbc4e6  t10k-images-idx3-ubyte.gz
+f7ae60f92e00ec6debd23a6088c31dbd2371eca3ffa0defaefb259924204aec6  t10k-labels-idx1-ubyte.gz
+440fcabf73cc546fa21475e81ea370265605f56be210a4024d2ca8f203523609  train-images-idx3-ubyte.gz
+3552534a0a558bbed6aed32b30c495cca23d567ec52cac8be1a0730e8010255c  train-labels-idx1-ubyte.gz
+EOT
+)
+
+mkdir -p "${download_dir}"
+cd "${download_dir}"
+echo "${mnist_sha256sums}" | cut -d' ' -f3 | xargs -I% wget -nv -nc "${base_url}/%"
+echo "${mnist_sha256sums}" | sha256sum --quiet --strict -c
--- a/test/mxnet/MNIST.d
+++ b/test/mxnet/MNIST.d
@@ -1,0 +1,438 @@
+/*******************************************************************************
+
+    Provides functionality to load the images and labels of the MNIST dataset
+
+    The MNIST dataset consists of labeled hand written digits. Each example is
+    a hand written digit `0` to `9` that is labeled with `0` to `9`. The hand
+    written digits are provided as 28x28 black and white images where a pixel's
+    intensity is encoded from 0 (white) to 255 (black).
+
+    The MNIST dataset provides a training and a testing dataset. The training
+    dataset consists of 60000 labeled images. The testing dataset consists of
+    10000 labeled images.
+
+    The dataset can be obtained from <http://yann.lecun.com/exdb/mnist/>.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH.
+
+    License:
+        Boost Software License Version 1.0.  See accompanying LICENSE.txt for
+        details, or <https://www.boost.org/LICENSE_1_0.txt>
+
+*******************************************************************************/
+
+module test.mxnet.MNIST;
+
+import core.stdc.string;
+import core.sys.posix.arpa.inet;
+
+import ocean.core.Enforce;
+import ocean.io.compress.ZlibStream;
+import ocean.io.device.File;
+import ocean.io.Path;
+import ocean.math.Math;
+import ocean.sys.Environment;
+import ocean.transition;
+
+
+///
+unittest
+{
+    // load the MNIST training dataset
+    auto mnist_training = trainingSet(datasetPath([]));
+    auto images = mnist_training.images;
+    auto labels = mnist_training.labels;
+
+    // dimensions of an image
+    auto image_rows = 28;
+    auto image_cols = 28;
+    assert(labels.length == images.length / (image_rows * image_cols));
+
+    // access the first image and its label
+    auto first_training_image = images[0 .. image_rows * image_cols];
+    auto first_training_label = labels[0];
+
+    // image is stored row wise with values between 0 (white) and 255 (black)
+    auto first_row = first_training_image[0 .. image_rows];
+}
+
+
+/*******************************************************************************
+
+    A Dataset consisting of images with their respective labels
+
+*******************************************************************************/
+
+struct Dataset
+{
+    /***************************************************************************
+
+        Images representing hand-written digits
+
+        Each image is of size 28 x 28 with its (pixel) values between 0 (white)
+        and 255 (black). The pixels are stored row wise and all the images'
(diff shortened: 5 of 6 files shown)


docker/build
```diff
@@ -17,3 +17,6 @@ apt-get update
 # Install dmxnet dependencies
 apt-get install -y \
     libmxnet
+
+# Download MNIST dataset
+$(dirname "$0")/download-mnist
```
joseph-wakeling-sociomantic: Note, this could be hardcoded to `/docker-tmp/download-mnist`.

leandro-lucarella-sociomantic (commented):
LGTM, although in term of travis resources usage, this will effectively download the DB on every test. You could use the [travis cache facilities](https://docs.travis-ci.com/user/caching/), which will download every time too, but at least I think it will play nicer tot he original DB servers.

joseph-wakeling-sociomantic:
Updated the last patch to add `zlib1g-dev` to the Dockerfile installed dependencies:

```diff
diff --git a/docker/build b/docker/build
index 96f5924..11a81ce 100755
--- a/docker/build
+++ b/docker/build
@@ -16,7 +16,8 @@ apt-get update
 
 # Install dmxnet dependencies
 apt-get install -y \
-    libmxnet
+    libmxnet \
+    zlib1g-dev
 
 # Download MNIST dataset
 $(dirname "$0")/download-mnist
```

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic so the dockerfile is regenerated with every test run? :-\

joseph-wakeling-sociomantic:
BTW a quick question about `.travis.yml`: do I explicitly need to add `DC=dmd-transitional` or `DVER=2` options for D2 builds?  I'm concerned that we are only running D1 tests of the library.

Build.mak
```diff
@@ -8,10 +8,15 @@ ifeq ($(DVER),1)
 	override DFLAGS += -v2 -v2=-static-arr-params -v2=-volatile
 endif
 
+test: override LDFLAGS += -lz
```
jens-mueller-sociomantic: This does not cover the `fasttest` target and the `integrationtest` test target. But we don't need it for the `unittest` target.

joseph-wakeling-sociomantic: Right, we only need it for the test suite itself.

jens-mueller-sociomantic:
> this will effectively download the DB on every test

This is no good because if the dataset web site goes down continuous integration won't work. I believe we need a solution for this.

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic agree, this makes our tests very flaky (especially given that if we run a lot of tests, we could be the cause of the flakiness).

leandro-lucarella-sociomantic:
Travis cache?

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic I'll look into the cacheing options.

And the D1/D2 build situation ... ?

leandro-lucarella-sociomantic:
What in particular?

jens-mueller-sociomantic:
Yes. The cache should be fine even though it download from S3. We just use it as another source of the data set.

Giving
```yaml
cache:
  directories:
  - /var/cache/mnist
```
a try is promising.

jens-mueller-sociomantic:
> BTW a quick question about `.travis.yml`: do I explicitly need to add `DC=dmd-transitional` or `DVER=2` options for D2 builds?  I'm concerned that we are only running D1 tests of the library.

It looks to me that we are only building and testing D1. I think `DVER=2` is fine but you also need to perform the conversion, don't you.

leandro-lucarella-sociomantic:
Bear in mind that stuff inside the docker image wont be saved, so you probably need to download the DB in plain travis, then you can add a `-v /var/cache/mnist` to `docker run` to use it.

joseph-wakeling-sociomantic:
https://docs.travis-ci.com/user/caching/#Things-not-to-cache

> Docker images are not cached, because we provision a brand new virtual machine for every build.

Right, AFAICS we need to add a `cache:` section to `.travis.yml` and specify what we need.  That actually looks like it makes things super easy, because we can just specify the one directory; I think it might even allow us to entirely avoid including the data download in the building of the docker image, as we can just rely on the makefile to do the job.

I'll push some fixup patches to try this out.

joseph-wakeling-sociomantic:
I've dropped the patch adding the MNIST data to the docker image, and instead updated the last patch to cache the MNIST data.  I've also made a few other suggested tweaks.

```diff
diff --git a/.travis.yml b/.travis.yml
index c9b6246..10ab969 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ sudo: required
 services:
     - docker
 
+cache:
+    directories:
+        - $(HOME)/.cache/mnist/
+
 # Disable automatic submodule fetching to prevent
 # it from being done recursively
 git:
diff --git a/Build.mak b/Build.mak
index bee9ce9..95cc5ff 100644
--- a/Build.mak
+++ b/Build.mak
@@ -8,8 +8,6 @@ ifeq ($(DVER),1)
        override DFLAGS += -v2 -v2=-static-arr-params -v2=-volatile
 endif
 
-test: override LDFLAGS += -lz
-
 # use NaiveEngine (non-threaded) MXNet engine since the default (threaded)
 # version dead locks
 %test: export MXNET_ENGINE_TYPE=NaiveEngine
@@ -18,5 +16,6 @@ test: override LDFLAGS += -lz
 download-mnist: $C/docker/download-mnist
        $(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
 
+$O/test-mxnet.stamp: override LDFLAGS += -lz
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
 $O/test-mxnet.stamp: download-mnist
diff --git a/Dockerfile.xenial b/Dockerfile.xenial
index 53c1e95..7c8d018 100644
--- a/Dockerfile.xenial
+++ b/Dockerfile.xenial
@@ -1,4 +1,3 @@
 FROM sociomantictsunami/dlang:v1-xenial
 COPY docker/ /docker-tmp
-ENV MNIST_DATA_DIR=/var/cache/mnist
 RUN /docker-tmp/build && rm -fr /docker-tmp
diff --git a/docker/build b/docker/build
index 11a81ce..6dc9046 100755
--- a/docker/build
+++ b/docker/build
@@ -18,6 +18,3 @@ apt-get update
 apt-get install -y \
     libmxnet \
     zlib1g-dev
-
-# Download MNIST dataset
-$(dirname "$0")/download-mnist
```

.travis.yml
```diff
@@ -7,6 +7,10 @@ sudo: required
 services:
     - docker
 
+cache:
+    directories:
+        - $(HOME)/.cache/mnist/
```
joseph-wakeling-sociomantic: Note, this is the default download location of the `download-mnist` script.  Seems better to use that if we can.

joseph-wakeling-sociomantic:
Looks to me like the caching is failing, which is unfortunate :-\  I'm going to try pushing a fixup patch that specifies `MNIST_DATA_DIR`.

docker/download-mnist
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
jens-mueller-sociomantic: I think this should be put outside `docker/` because we need to run it outside of docker. `script/` seems to be the right place.

joseph-wakeling-sociomantic: IIRC with the original implementation, there were some difficulties accessing it when running the docker image build, except if it was in the `docker/` dir.  However, since we're not baking the data into the docker image anyway, we can move it (and come back to this problem if we ever do need to bake the data into the image).

leandro-lucarella-sociomantic:
I can't find where the mnist DB is actually being downloaded, but as I see it you need to add another command to the `install` stage in Travis, you could even run the download right inside docker not to worry about setting up dependencies in Travis itself:
```yml

env:
    global:
        - BEAVER_DOCKER_OPTS="-v $HOME/.cache/mnist:$HOME/.cache/mnist"

install:
    - beaver install
    - beaver make download-mnist
```

You will actually have to wait until [this beaver PR](https://github.com/sociomantic-tsunami/beaver/pull/16) to be merge to be able to pass extra options to `beaver run/make` though.

Also, you will have to introduce some way to not re-download mnist if it's already present in the cache directory (I didn't see any checks like this in the download script or make, but maybe I missed it).

joseph-wakeling-sociomantic:
Is it necessary to invoke `make download-mnist` in the `install:` section?  Note that it's a dependency of the `test-mxnet.stamp` target, so will be invoked anyway when the test suite is run.

joseph-wakeling-sociomantic:
Rebased on master.  The `download-mnist` directory is now placed in the `script` directory, and `.travis.yml` has been updated to include the `BEAVER_DOCKER_OPTS`.

Of course, the latter will not work right now, but it should be possible to at least review the general setup and come back to this once we are in a position to update `beaver`.

joseph-wakeling-sociomantic:
Marking as blocked until `beaver` has been updated.

joseph-wakeling-sociomantic:
Woah, it looks like this is working already.  I'll try a rebuild and see how it copes.

joseph-wakeling-sociomantic:
Success!! :-D

joseph-wakeling-sociomantic:
Un-blocking, as it looks like this does what we need/want.

leandro-lucarella-sociomantic:
Ah, right, because you are storing it in `$HOME` and I think travis sets the HOME as the project's dir, which is already shared between docker and the host

leandro-lucarella-sociomantic:
OTOH, I would download mnist as part of the install steps, if that fails is not really an error in the application, so the build shouldn't be marked as failed, but as "errored".

leandro-lucarella-sociomantic:
> I think travis sets the HOME as the project's dir

This is false, but beaver is explicitly sharing the `$HOME` between the host and docker, so we should be fine by using HOME, no need to make an explicit share in here. :+1:

joseph-wakeling-sociomantic:
I'll try cleaning the cache and retrying without the `BEAVER_DOCKER_OPTS` to see if it makes a difference.

joseph-wakeling-sociomantic:
(It didn't seem to work before I added `BEAVER_DOCKER_OPTS`, even though I was using `$HOME/.cache/mnist` for the download dir.)

joseph-wakeling-sociomantic:
> OTOH, I would download mnist as part of the install steps, if that fails is not really an error in the application, so the build shouldn't be marked as failed, but as "errored".

Fair enough, I'll tweak.

joseph-wakeling-sociomantic:
OK, looking good.  I'll squash and then hopefully we can have a last round of review.

joseph-wakeling-sociomantic:
Squashed, and rewrote the last commit message.  This is ready for a last round of review.

(Can everyone please check that the commit message writeups make sense to them?)

jens-mueller-sociomantic:
In the second commit message it says:

> In general however this should not be necessary, since the defaults will
work fine for local testing, while CI setup can ensure `MNIST_DATA_DIR`
is defined if needed in that context (e.g. by explicitly setting it in
the docker image, or in the config for CI environment variables).

We currently do not set up `MNIST_DATA_DIR` for travis at all. So setting `MNIST_DATA_DIR` in the environment is useful but actually not needed for travis CI. So maybe rephrase this as a general option for configuration but not tie it so much to CI.

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic agreed, we don't do this, but it's a note of what _can_ be done rather than what _will_ be done.

joseph-wakeling-sociomantic:
Note the "... if needed in that context" remark.

leandro-lucarella-sociomantic (commented):
As far as I can see this is still lacking the check to see if mnist is already present, so it will be downloaded every time. Or am I missing something?

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic that's in the `download-mnist` script itself, or rather in how `wget` is invoked with the `-nc` (`--no-clobber`) flag.

jens-mueller-sociomantic (approved):

PR #3, opened 2017-08-03
Add D2 builds to CI

joseph-wakeling-sociomantic:
The original `matrix:` setup only included D1 builds.  This patch should ensure that both D1 and D2 builds are carried out by Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,13 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
+        - BEAVER_DOCKER_VARS="DC F"
     matrix:
-        - DIST=xenial F=production
-        - DIST=xenial F=devel
+        - DC=dmd1 DIST=xenial F=production
+        - DC=dmd1 DIST=xenial F=devel
+        - DC=dmd-transitional DIST=xenial F=production
+        - DC=dmd-transitional DIST=xenial F=devel
 
 install: beaver install
 
-script: beaver make test all
+script: beaver run ci/run.sh
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+set -eux
+
+DVER=1
+if test "$DC" != dmd1; then
+	DVER=2
+fi
+
+export DC DVER
+
+if test "$DC" != dmd1; then
+	make -r d2conv
+fi
+
+make -r all
+make -r test


joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic I'm sorry to keep bothering you regarding dmxnet stuff, but can you advise on this?  It looks like the D compiler and verbosity settings are not being taken into account at all by the build.

jens-mueller-sociomantic:
@joseph-wakeling-sociomantic How do you know the D compiler is not used?

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic I am not certain (given that it's not a verbose build) but two things are striking:

* the build is not verbose despite explicitly setting `V=1` (suggesting that some environment variables are being ignored)

* I see no sign that `d1to2fix` is being run in the builds using `dmd-transitional`.

The original CI implementation (that used custom shell scripts rather than `beaver`) dealt with all this manually, so that if one specified anything other than `dmd1` as the compiler, D2 conversion would be applied.  But it doesn't look like `beaver` is handling this.

jens-mueller-sociomantic:
According to the documentation something like

```yaml
env:
    global:
        - BEAVER_DOCKER_VARS="DC F"
    matrix:
        - DIST=xenial DC=dmd1 F=production
        - DIST=xenial DC=dmd1 F=devel
        - DIST=xenial DC=dmd-transitional F=production
        - DIST=xenial DC=dmd-transitional F=devel

install: beaver docker build -f docker/Dockerfile .
```
should do it where setting `BEAVER_DOCKER_VARS` is the important bit.

> If you need to pass extra variables to docker run you can do it by using the `BEAVER_DOCKER_VARS` environment variable.

from <https://github.com/sociomantic-tsunami/beaver/blob/master/README.md>

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic yup, sounds good.  I've pushed a fixup patch implementing this, let's see how it goes.

joseph-wakeling-sociomantic:
Squashed down to a single commit and rewrote the commit message.

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic SLGTY? :-)

.travis.yml
```diff
@@ -19,10 +19,14 @@ env:
...
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
+        - BEAVER_DOCKER_VARS="DC F"
     matrix:
-        - DIST=xenial F=production
-        - DIST=xenial F=devel
+        - DC=dmd1 DIST=xenial F=production
+        - DC=dmd1 DIST=xenial F=devel
+        - DC=dmd-transitional DIST=xenial F=production
+        - DC=dmd-transitional DIST=xenial F=devel
```
nemanja-boric-sociomantic: You don't need trusty, fair enough.

joseph-wakeling-sociomantic:
Just made one minor tweak:

```diff
diff --git a/.travis.yml b/.travis.yml
index 49f0046..5e5479a 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,5 +28,4 @@ env:
 
 install: beaver install
 
-script:
-    - beaver run ci/run.sh
+script: beaver run ci/run.sh
```

... since we never had a multi-entry `script:` location before (it was a leftover from some transitional stuff mid-fixup-commits).

.travis.yml
joseph-wakeling-sociomantic: Yea, we decided to exclude `trusty` from the beginning.  Not that it won't work, I think, but I don't think we have any need or reason to want to test there.

PR #7, opened 2017-08-24
Update beaver from v0.1.4 to v0.2.1

joseph-wakeling-sociomantic:
This new beaver release brings with it some useful new features for creating and using dockerfile definitions.

* submodules/beaver v0.1.4(3c827c9)...v0.2.1(570562d) (11 commits)
  > bintray: Fix upload command-line arguments
  > Add a simple Makefile for easy local testing
  > dlang: Make install install DMD automatically
  > install: Use gen-dockerfile if beaver.Dockerfile is present
  > docker: Add new gen-dockerfile command
  (...)


joseph-wakeling-sociomantic:
Blocking this until we can update `makd`.  That should enable us to replace the `test` dir with `integrationtest` and from there it should be possible to make this upgrade (given that the `test` package name clashes with the `test` function template).

We _could_ alternatively specify a precise `DMD` version, but I'd rather avoid that if possible, as that most likely creates maintenance issues of its own.

joseph-wakeling-sociomantic:
Rebased on master and updated to v0.2.1.  I've pushed the update patch alone first in order to ensure that it gets verified in its own right.

alaksiej-stankievicx-dunnhumby (dismissed):

joseph-wakeling-sociomantic:
Restored the dockerfile switch.  I may follow up on this with a further patch to make `.travis.yml` changes.

joseph-wakeling-sociomantic:
Minor correction to description of 2nd commit.

joseph-wakeling-sociomantic:
Added a 3rd patch to update Travis CI config to use new `beaver dlang` commands

joseph-wakeling-sociomantic:
Added a temporary patch to ensure `V=1` for all builds.  (I might consider squashing this into the preceding patch if it looks useful.)

joseph-wakeling-sociomantic:
OK, temporary verbose build looks good, so dropping that patch.  Verbose output is a bit overwhelming, so not so useful.

However, it would be useful for D builds to be able to explicitly see the D compiler and D build flags in use (and not other D output).

joseph-wakeling-sociomantic:
Added one last patch to remove our old CI run script, since this should no longer be needed.

joseph-wakeling-sociomantic:
OK, I think this is now ready for review.  @leandro-lucarella-sociomantic @nemanja-boric-sociomantic it would be good to have some core-team/beaver dev oversight on this to make sure it's all being done in line with intentions and expectations ;-)

nemanja-boric-sociomantic:
This looks good, except I don't know what kind of an impact doing `sudo apt-get upgrade` has in the `docker/build` script and different compilers available.

joseph-wakeling-sociomantic:
@nemanja-boric-sociomantic it should mean we're always testing with the latest compiler and other dependency versions.  I did consider pinning things or requesting specific compiler versions, but TBH it seems better to avoid that if possible, for reasons we've already discussed.

Any thoughts on the issue of being able to validate from the build log that the correct compiler and build flags are being used?

nemanja-boric-sociomantic:
I think that's a fair requirement, I'm not sure if/how that would be possible. @leandro-lucarella-sociomantic may tell you more.

.travis.yml
```diff
@@ -23,15 +23,15 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
-        - BEAVER_DOCKER_VARS="DC F"
+        - BEAVER_DOCKER_VARS="F"
```
joseph-wakeling-sociomantic, line 26: Is it still necessary to pass `F` explicitly to `BEAVER_DOCKER_VARS`?  IIUC `DC` is now correctly autogenerated by either `dlang install` or `dlang make`.

joseph-wakeling-sociomantic: I'm assuming that it's reasonable that this should stay a manual requirement, since there's no way that beaver can automagically know that makd is being used to manage the build (even if makd is a submodule).

nemanja-boric-sociomantic: This is what I think a well, but ocean is not doing it, so probably ocean is wrong; https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/.travis.yml

joseph-wakeling-sociomantic: Tell you what.  I'll throw together a `V=1` here together with removing the passing of `F` and see what happens.  Then we can revert once we have clarity.

joseph-wakeling-sociomantic:
Added a temporary commit just to verify if we do need to pass `F` in `BEAVER_DOCKER_VARS`.

nemanja-boric-sociomantic:
Does the travis have problems testing this?

joseph-wakeling-sociomantic:
Don't know.  It may just be slow picking things up (IIRC it can be like that sometimes if it's busy).

joseph-wakeling-sociomantic:
It seems to have picked up on it now, though.  Did you do anything?  (I did nothing, apart from browsing the travis dashboard for dmxnet...)

nemanja-boric-sociomantic:
No, I didn't even open the travis dashboard.

joseph-wakeling-sociomantic:
I think it was probably GitHub's hooks that were slow, as AFAICS the actual build ran a while ago.

Anyway, these are the relevant "production" builds:
https://travis-ci.org/sociomantic-tsunami/dmxnet/jobs/270047319
https://travis-ci.org/sociomantic-tsunami/dmxnet/jobs/270047321

As you can see from the D flags, they don't use `-O -inline`, so I think we can reasonably say that `F` is a required parameter ;-)  I'll just push one more fixup patch to verify that things go right again once we have it.

nemanja-boric-sociomantic:
> As you can see from the D flags, they don't use -O -inline, so I think we can reasonably say that F is a required parameter ;-) I'll just push one more fixup patch to verify that things go right again once we have it.

Looks like it!

nemanja-boric-sociomantic:
Though, look into the ocean's production build: https://travis-ci.org/sociomantic-tsunami/ocean/jobs/269935163

It puts the data inside `build/production/...`

nemanja-boric-sociomantic:
Ignore my last comment it was the old build

joseph-wakeling-sociomantic:
With `F` restored, the production builds have the right flags:
https://travis-ci.org/sociomantic-tsunami/dmxnet/jobs/270051782
https://travis-ci.org/sociomantic-tsunami/dmxnet/jobs/270051784

... so once those builds are complete, I'll drop my dodgy test patches and re-push ;-)

joseph-wakeling-sociomantic:
Dropped the 2 test patches.  @nemanja-boric-sociomantic do you see it as more future proof to define `BEAVER_DOCKER_VARS` per `script:` entry (cf. your ocean fix)?  It seems fine for now but obviously at some point we might change what we build here.

nemanja-boric-sociomantic:
IMHO, having non-global `BEAVER_DOCKER_VARS` makes more sense.

joseph-wakeling-sociomantic:
OK, great.  All still LGTY? :-)

nemanja-boric-sociomantic (approved):
Still LGTM

PR #9, opened 2017-08-30
Update makd from v1.10.0 to v2.0.1

joseph-wakeling-sociomantic:
~~This patch is a test to validate that the v2.0.0 release candidate works for dmxnet.  It should be amended once v2.0.0 final is released.  @leandro-lucarella-sociomantic FYI since you asked for tests ;-)~~

With v2.0.0 (and v2.0.1!) now released, we can transition this to a final patch.

Since dmxnet defines no packages, the number of transitional changes required for this major version update is fairly small: only adding a `version (UnitTest) {} else` exclusion for the integration-test `main()` function, and adding the `-lz` dependency to unittests (since unittests now also run for integration-test modules).

* submodules/makd v1.10.0(ab31113)...v2.0.1(d735c1d) (20 commits)
  > Fix calls to `find`
  > Clear release notes after release
  > Don't print errors if there is no directory to find
  > Document the `find_files` function
  > relnotes: Mention more potential problems
  (...)


joseph-wakeling-sociomantic:
Updated to v2.0.1 stable release.  This is now ready for review.

joseph-wakeling-sociomantic:
Rebased on master.  It looks like this upgrade interferes with the `DFLAGS` override for the `$O/test-mxnet.stamp` target, so blocking again.  @leandro-lucarella-sociomantic can you suggest what might be going wrong here?

@nemanja-boric-sociomantic pointed to https://github.com/sociomantic-tsunami/makd/blob/v2.0.1/Makd.mak#L609-L616 as a potential source of this issue.

joseph-wakeling-sociomantic:
Rebased on latest v0.x.x branch, which includes fixes to the test-suite `Build.mak` targets that should address the problems observed.  However, I'm still seeing no evidence of `-debug=MXNetHandleManualFree` working, even though it does look like it's now passed to the build of the integration tests.

joseph-wakeling-sociomantic:
Blocking until this concern is resolved.

jens-mueller-sociomantic:
> it does look like it's now passed to the build of the integration tests

I see no `-debug=MXNetHandleManualFree` being passed when building.

joseph-wakeling-sociomantic:
> I see no `-debug=MXNetHandleManualFree` being passed when building.

It is being passed, just appears to have no effect now.  I'll check if there is a change to _where_ it's being passed and whether this makes a difference.

Adding `-debug=MXNetHandle` to the same DFLAGS override results in correct behaviour for that debug message.

joseph-wakeling-sociomantic:
Ah, all the cases of non-freed handles in the integration test code are actually in _unittests_ within the integration tests.  See https://github.com/sociomantic-tsunami/dmxnet/pull/14 for details.

So, this is correct behaviour: those cases will now show up in `allunittests` rather than during the integration-test run.

joseph-wakeling-sociomantic:
OK, this can be considered ready for other review.

jens-mueller-sociomantic:
> So, this is correct behaviour: those cases will now show up in `allunittests` rather than during the integration-test run.

But we don't pass `-debug=MXNetHandleManualFree` for `unittest`s. So with this change we will only non-freed handles created inside the integration test.

joseph-wakeling-sociomantic:
Yes, and I think that's fine.  We can always do some local tests before/after this update with `unittest` builds, just to validate that the behaviour changes as we expect.

joseph-wakeling-sociomantic:
Rebased on latest v0.x.x branch.

jens-mueller-sociomantic (approved):

leandro-lucarella-sociomantic:
> Rebased on master. It looks like this upgrade interferes with the DFLAGS override for the $O/test-mxnet.stamp target, so blocking again. @leandro-lucarella-sociomantic can you suggest what might be going wrong here?
> 
> @nemanja-boric-sociomantic pointed to https://github.com/sociomantic-tsunami/makd/blob/v2.0.1/Makd.mak#L609-L616 as a potential source of this issue.

This is already resolved (was the wrong make target), right?

joseph-wakeling-sociomantic:
> This is already resolved (was the wrong make target), right?

Right :-)

PR #11, opened 2017-09-12
Add debugging help when freeing handles in the destructor

jens-mueller-sociomantic:
Freeing non-null handles in the destructor is considered to be a
programming error because they should be freed by an explicit call to
`freeHandle`. This change adds debugging output when a non-null handle
is freed inside the destructor. Because relying on the GC for freeing a
handle in time is not reliable. It is reliable for `delete` or `destroy`
but we prefer an explicit call to `freeHandle` prior to a manual
`delete` or `destroy`. The policy is to always free the handle manually
using `freeHandle`.

This debugging diagnostics is enabled for `MXNetHandleManualFree` debug
builds.

### DIFF
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -432,14 +432,16 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
         {
             if (this.exists())
             {
-                // we require manual resource management by calling
-                // `freeHandle` to free MXNet resources explicitly and never
-                // rely on the destructor
+                // Relying on the GC calling this destructor to free the handle
+                // is problematic since it may not be in-time. For this reason,
+                // we recommend users of `MXNetHandle` to free MXNet resources
+                // by calling `freeHandle` manually, and never relying on the
+                // destructor.
                 //
-                // relying on the GC calling this destructor to free the handle
-                // is problematic since it may not be in-time
-                // performing a manual `delete` or `destroy` is technically
-                // fine but forbidden by policy
+                // If this policy is followed, then the handle should already
+                // have been freed by the time the destructor is initiated.
+                // This debug check should help in tracking down any handles
+                // for which this has not been done.
                 sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
                         "Non-null handle {:x} in MXNetHandle destructor!\n",
                         this.c_api_handle);
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -428,6 +428,23 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
 
     public ~this ()
     {
+        debug (MXNetHandleManualFree)
+        {
+            if (this.exists())
+            {
+                // we require manual resource management by calling
+                // `freeHandle` to free MXNet resources explicitly and never
+                // rely on the destructor
+                //
+                // relying on the GC calling this destructor to free the handle
+                // is problematic since it may not be in-time
+                // performing a manual `delete` or `destroy` is technically
+                // fine but forbidden by policy
+                sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
+                        "Non-null handle {:x} in MXNetHandle destructor!\n",
+                        this.c_api_handle);
+            }
+        }
         this.freeHandle();
     }
 }


src/mxnet/Handle.d
```diff
@@ -428,6 +428,22 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+            {
+                // we require manual resource management by calling
+                // `freeHandle` to free MXNet resources explicitly and never
+                // rely on the destructor
+                //
+                // relying on the GC calling this destructor to free the handle
+                // is problematic since it may not be in-time
+                // performing a manual `delete` or `destroy` is technically
+                // fine but forbidden by policy
+                const msg = "Non-null handle in MXNetHandle destructor!\n";
```
joseph-wakeling-sociomantic: You might want to include the handle pointer if you can.  In combination with the other debug option, it'll probably make it easier to trace the point of creation of the problematic handle.

joseph-wakeling-sociomantic (approved):
LGTM, but you might want to tweak as suggested for even more useful debug output.

jens-mueller-sociomantic:
Note we have code in this file using `snformat` with a buffer. I find the
following cleaner since it avoids the buffer. But I wonder why the other code
was written using `snformat` instead of `sformat`.

```diff
diff --git a/src/mxnet/Handle.d b/src/mxnet/Handle.d
index aab35be..16b0439 100644
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -440,8 +440,9 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
                 // is problematic since it may not be in-time
                 // performing a manual `delete` or `destroy` is technically
                 // fine but forbidden by policy
-                const msg = "Non-null handle in MXNetHandle destructor!\n";
-                write(STDERR_FILENO, msg.ptr, msg.length);
+                sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
+                        "Non-null handle {:x} in MXNetHandle destructor!\n",
+                        this.c_api_handle);
             }
         }
         this.freeHandle();
```

jens-mueller-sociomantic:
A problem here might be the GC allocation of the delegate in D2. I need to check whether that actually GC-allocates.

joseph-wakeling-sociomantic:
`snformat` with a fixed-size buffer was indeed used in order to avoid GC.  But I'm fine with switching to the sink-delegate-based approach, as long as the delegate doesn't GC-allocate.

Just note that you have to be a _little_ careful with `sformat` because of the contradictory meanings of its overloads (the sink-based ones are fine, but the buffer-based one _appends_ to the buffer instead of writing into it).

src/mxnet/Handle.d
```diff
@@ -428,6 +428,23 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+            if (this.exists())
+            {
+                // we require manual resource management by calling
+                // `freeHandle` to free MXNet resources explicitly and never
+                // rely on the destructor
+                //
+                // relying on the GC calling this destructor to free the handle
+                // is problematic since it may not be in-time
+                // performing a manual `delete` or `destroy` is technically
+                // fine but forbidden by policy
```
joseph-wakeling-sociomantic: Is the problem that it's forbidden by policy, or that it's not guaranteed to call the destructor?

jens-mueller-sociomantic: Forbidden by policy to remove any false positives from this debugging diagnostics. How would you rephrase?

joseph-wakeling-sociomantic: Well IIRC the _real_ reason for enforcing manual calling of `freeHandle` is not so much for `Handle` itself (we could use `scope class` to guarantee the GC cleans up), but because we can't rely on wrapper classes like `Symbol` or `NDArray` to ensure the `Handle` destructor is invoked (even if they are `scope` themselves).  No ... ?

jens-mueller-sociomantic: True. We also need it for the existing wrappers. But I'd like to avoid using `delete` or `destroy` on any `MXNetHandle`. I mean technically a wrapper could use `delete` or `destroy` inside a member function and not `MXNetHandle.freeHandle`. Which gives a simple rule whenever this diagnostic occurs: Never rely on the destructor. Call always `freeHandle` manually.
Though probably unlikely, a wrapper might use `delete` or `destroy` to ensure the `MXNetHandle` destructor is called. I'd like to forbid this by policy.

joseph-wakeling-sociomantic: Part of what's making me uncomfortable is the claim that `Performing a manual delete or destroy is technically fine`.

`delete` has the obvious problem of not guaranteeing that `~this` will be called in a timely fashion.  `destroy` IIRC is inconsistent in how it's implemented between D1 and D2.  So I'm not sure either of them _is_ "technically fine" in context.

jens-mueller-sociomantic: It says
> If the UnaryExpression is a class object reference, and there is a destructor for that class, the destructor is called for that object instance.

on https://dlang.org/spec/expression.html#DeleteExpression

I read this as the destructor is called immediately but maybe that is not the case.

I don't know of any `destroy` in D1. I only know this one https://dlang.org/phobos/object.html#.destroy in Phobos. This `destroy` is like a `delete` minus the GC related operations performed by `delete`.

joseph-wakeling-sociomantic: > I don't know of any `destroy` in D1.

I may be misremembering, I thought there was a method somewhere in `Object`, or elsewhere in the tango runtime.

Anyway, I think the broader point is, if we could reasonably rely on `destroy` in both D1 and D2, we'd probably use it (better to have a standard language mechanism than a custom-written method).  It's the fact that we _don't_ have that option reliably available to us that motivates us to insist on manual calling of `freeHandle` from code external to `Handle` itself.

I'd suggest dropping remarks about policy.  How about this as a comment text?

```D
// Relying on the GC calling this destructor to free the handle
// is problematic since it may not be in-time.  For this reason,
// we recommend users of `Handle` (or classes that wrap it) to
// free MXNet resources by calling `freeHandle` manually, and
// never relying on the destructor.
//
// If this policy is followed, then the handle should already have
// been freed by the time the destructor is initiated.  This debug
// check should help in tracking down any handles for which this
// has not been done.
```

In other words we don't _mandate_ any particular user choice, but we make a recommendation and give clear advice on how to ensure that recommendation is followed effectively.

src/mxnet/Handle.d
```diff
@@ -428,6 +428,23 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+                sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
```
joseph-wakeling-sociomantic: @mathias-lang-sociomantic I wonder if you might comment on whether this delegate is likely to GC-allocate or not, and if so, how to address that?

mathias-lang-sociomantic: Why would the delegate allocate ?

jens-mueller-sociomantic: What about the delegate itself. Is it GC-allocated in D2 or is that optimized away?

src/mxnet/Handle.d
jens-mueller-sociomantic: I pushed a fixup using a slight modification of the suggested comment.

src/mxnet/Handle.d
jens-mueller-sociomantic: For the record: In D2 the delegate will be passed using `scope`. This is a result of `d1to2fix` which will add `scope` for the delegate parameter. The compiler won't allocate here since the delegate is anonymous (only defined for the call). Moving the definition of the delegate outside using `scope dg = (cstring str) { write(STDERR_FILENO, str.ptr, str.length);` would also work but this usage of `scope` is deprecated in D2 (see https://dlang.org/deprecate.html#scope%20for%20allocating%20classes%20on%20the%20stack).

mathias-lang-sociomantic: Well it's deprecated but it's not going anywhere, trust me ;)

joseph-wakeling-sociomantic (commented):
LGTM in itself, so feel free to squash :-)

Regarding the delegate: it looks to me like this `sformat`-with-sink approach might be useful in a few places, so it might be worth taking it outside the class as a private helper function, say `debugMessageSink`.  What do you think?

jens-mueller-sociomantic:
Yes. This pattern occurs more than once. I agree that we should have a function for it.

joseph-wakeling-sociomantic:
Force-pushed to reset `v0.x.x` branch, since this should not have been merged.

joseph-wakeling-sociomantic:
Doesn't look like we can re-open this PR so I would suggest submitting a new one.

jens-mueller-sociomantic:
Too bad this one says merged now. But your comment should be enough to indicate that this did not end up in `v0.x.x`.

PR #12, opened 2017-09-13
Add debugging help when freeing handles in the destructor

jens-mueller-sociomantic:
Freeing non-null handles in the destructor is considered to be a
programming error because they should be freed by an explicit call to
`freeHandle`. This change adds debugging output when a non-null handle
is freed inside the destructor. Because relying on the GC for freeing a
handle in time is not reliable. It is reliable for `delete` or `destroy`
but we prefer an explicit call to `freeHandle` prior to a manual
`delete` or `destroy`. The policy is to always free the handle manually
using `freeHandle`.

This debugging diagnostics is enabled for `MXNetHandleManualFree` debug
builds.


jens-mueller-sociomantic:
This is the proper PR for sociomantic-tsunami/dmxnet#11 which was wrongly merged and removed from the history.

src/mxnet/Handle.d
```diff
@@ -428,6 +428,25 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+                // is problematic since it may not be in-time. For this reason,
+                // we recommend users of `MXNetHandle` to free MXNet resources
+                // by calling `freeHandle` manually, and never relying on the
+                // destructor.
+                //
+                // If this policy is followed, then the handle should already
+                // have been freed by the time the destructor is initiated.
+                // This debug check should help in tracking down any handles
+                // for which this has not been done.
+                sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
```
joseph-wakeling-sociomantic, line 445: Would you prefer we take in this patch as is, and follow up on separating out the sink, or shall we do it now as a tweak to this PR?  (If the latter, I'd suggest adding a first patch to introduce it.)

jens-mueller-sociomantic: I want to leave it for another PR.

joseph-wakeling-sociomantic: Well, in principle LGTM.  It's interesting to note, though, that this shows up during the integration tests on a number of occasions, and it's not obvious to me why, given that AFAICT we manually free everything.  Can you clarify why this is happening?  I'd just like to be sure that we are checking for what we think we are.

joseph-wakeling-sociomantic (approved):
Since fixing the missing test-suite `freeHandle` calls is a planned follow-up patch, LGTM :-)

Feel free to take in at your convenience as soon as the CI has gone green.

src/mxnet/Handle.d
jens-mueller-sociomantic: There are places where we miss the manual free. In a sense it is good that you found this because this shows that is debugging check is useful.

PR #13, opened 2017-09-13
Report missing manual free when running integration tests

jens-mueller-sociomantic:
For integration tests any handle that is not freed by having called
`freeHandle` will be reported. These checks are enabled by adding the
recently introduced debugging checks `MXNetHandleManualFree` when
building the integration tests. This helps to track down missing calls
to `freeHandle` when running these tests during development.


joseph-wakeling-sociomantic (approved):
LGTM, let's see what CI has to say :-)

PR #14, opened 2017-09-13
Free all non-null MXNet handles explicitly

jens-mueller-sociomantic:
Currently we have four places where we failed to explicitly free MXNet
non-null handles. We relied on the destructor being called. This change
adds explicit freeing of those four handles. All handles are now
explicitly freed.


joseph-wakeling-sociomantic (approved):
LGTM, but I'd like to hold off merging this until we've resolved the issues identified with the makd update.

joseph-wakeling-sociomantic:
One thing I'd like to clearly understand before merging this: when running integration tests, I sometimes get multiple debug messages before the test suite has even really got under way.  Example:

```
   run build/devel/tmp/test-mxnet
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
Non-null handle 20e0270 in MXNetHandle destructor!
[19:28:11] src/engine/engine.cc:36: MXNet start using engine: NaiveEngine
Non-null handle 20e0ae0 in MXNetHandle destructor!
MXNet handles in use at start: 1
Non-null handle 20e0f00 in MXNetHandle destructor!
Percentage of correctly predicted digits on MNIST testing: 0.9193%
Calculated in 1.214604 seconds
MXNet handles in use at finish: 10
MXNet handles in use at exit: 0
```

What's the reason for the _multiple_ messages about the same pointer?

jens-mueller-sociomantic:
I'll guess that is because the memory is reused. I mean the handle is just a pointer to some memory. If you're done with some memory that memory might be reused.

jens-mueller-sociomantic:
```
   run build/devel/tmp/test-mxnet
[12:39:38] src/engine/engine.cc:36: MXNet start using engine: NaiveEngine
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 17bd1b0 in MXNetHandle destructor!
Non-null handle 184aa70 in MXNetHandle destructor!
Non-null handle 17bc580 in MXNetHandle destructor!
Non-null handle 17bc580 in MXNetHandle destructor!
MXNet handles in use at start: 1
Non-null handle 17bc580 in MXNetHandle destructor!
Percentage of correctly predicted digits on MNIST testing: 0.9193%
Calculated in 1.171698 seconds
MXNet handles in use at finish: 10
MXNet handles in use at exit: 0
```

Let me do the counting here. We have to account for twelve reportings of non-null handles. `MXNetHandle` is instantiated five times (once in `Symbol`, once in `NDArray`, once in `Executor` and twice inside the unittests of `MXNetHandle` itself). So the first introduced `scope(exit)` inside `MXNetHandle` removes five reportings. Same for the second introduced `scope(exit)`. In total adding the missing `scope(exit)` inside `MXNetHandle` explains for the ten of the twelve reportings. The remaining two directly relate to the introduced `scope(exit)` inside `Symbol` and `Executor`, respectively.
In summary even though this PR adds only four `scope(exit)` all cases are addressed twelve cases are addressed since `MXNetHandle` is instantiated five times. Hence the two fixes inside `MXNetHandle` account for ten reportings instead of two.

joseph-wakeling-sociomantic:
OK.  Feel free to take in at your own convenience (in the circumstances, probably better to do it before the makd update rather than after).

joseph-wakeling-sociomantic:
I would suggest that after this is merged we add standalone unittests for all our data structures that validate that the destructor does still work properly if it needs to, but these should be one-case-per-data-structure rather than triggered for each _instantiation_ of that data structure.

PR #15, opened 2017-09-13
Refactor error reporting used for debugging

jens-mueller-sociomantic:
In five places errors are reported which may run in the context of
object destruction. Since it is not allowed to allocate during
destruction a buffer together with `snformat` is used in four of those
places. The other case uses `sformat` and a sink which is identical to
the implementation of `debugReport`. This change replaces this error
reporting by calling the recently introduced function `debugReport`
which also performs no allocation for these cases. This patch also
changes two usages of `Stderr.formatln` inside `debug` for consistency.


src/mxnet/Handle.d
```diff
@@ -442,16 +428,42 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+
+private void stderrWritef (Args...) (cstring fmt, Args args)
+{
+    sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
+            fmt, args);
+}
+
+unittest
+{
+    testNoAlloc(stderrWritef("the number {}\n", 1));
```
jens-mueller-sociomantic: How to test this properly? As this will write to `stderr` during testing.

src/mxnet/Handle.d
```diff
@@ -442,16 +428,42 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
+    function is safe to be executed inside a destructor or other places where
+    no allocation is supposed to happen.
+
+    Params:
+        fmt = format string to use
+        args = variadic arguments to format according to fmt
+
+*******************************************************************************/
+
+private void stderrWritef (Args...) (cstring fmt, Args args)
```
joseph-wakeling-sociomantic: Interesting choice of how to design this, I like it! :-)

I would suggest a rename in line with its _purpose_ rather than (as now) what it _does_.  In other words, call it something like `debugReport` or `reportDebugMessage`, and document accordingly.

src/mxnet/Handle.d
joseph-wakeling-sociomantic: I would suggest making the message here a little more informative, otherwise it will just look weird in the unittest build log.

src/mxnet/Handle.d
```diff
@@ -380,39 +378,27 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
         {
             // we cannot throw (even an Error) from here, since this
-            // method may be called by the destructor, so we use C
-            // stdlib to output an error message before aborting
-            const istring msg = "Could not free MXNet handle using " ~
-                                identifier!(FreeHandleFunction) ~ "@" ~
-                                __FILE__ ~ ":" ~ __LINE__.stringof ~ "\n";
-            write(STDERR_FILENO, msg.ptr, msg.length);
+            // method may be called by the destructor
+            stderrWritef("Could not free MXNet handle using {}@{}:{}\n" ~
```
joseph-wakeling-sociomantic: The trailing concatenation here looks wrong ... ?

src/mxnet/Handle.d
```diff
@@ -380,39 +378,27 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
             {
                 // we cannot throw (even an Error) from here, since this
-                // method may be called by the destructor, so we use the
-                // C stdlib to output an error message before aborting
-                const istring negative_handles_msg =
-                    "Allocated handle count is negative!\n";
-                write(STDERR_FILENO,
-                      negative_handles_msg.ptr, negative_handles_msg.length);
+                // method may be called by the destructor
+                stderrWritef("Allocated handle count is negative!\n");
```
joseph-wakeling-sociomantic: Since you can format the output now, how about including the handlecount value in the formatted message?

src/mxnet/Handle.d
```diff
@@ -380,39 +378,27 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
             debug (MXNetHandle)
             {
-                // using fixed stack buffer, `snformat` and `write` should
-                // avoid allocation, which (whether GC or not) might cause
-                // problems if this method is called from a destructor
-                char[80] buf; // enough space for message, `hash_t` and `long`
-                auto free_handle_msg =
-                    snformat(buf, "Freeing handle: {:x} ({} still allocated)\n",
+                // should avoid allocation, which (whether GC or not) might
+                // cause problems if this method is called from a destructor
```
joseph-wakeling-sociomantic: I'm not sure whether we need these repeated comments about allocation any more.  Maybe better to clarify this in documentation of the new function (i.e. that it is designed to be used to output debug messages in circumstances where allocation and throwing is not permitted).

src/mxnet/Handle.d
```diff
@@ -442,16 +428,42 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
...
     }
 }
 
 
+/******************************************************************************
+
+    Writes a formatted string to standard error
+
+    The only allocations this function performs are the ones necessary for
+    string conversions of the arguments. If no allocations are performed this
```
joseph-wakeling-sociomantic: This comment about `The only allocations ...` might be phrased a bit better I think.  It's important to be clear that this refers to GC allocation only, and that it is expected there will be _no_ GC allocation unless an argument _cannot_ be formatted without it.

In practice, this means it should be nogc for any built-in type (which is the entirety of our use-case).

joseph-wakeling-sociomantic (changes requested):
Looks pretty good.  There look like a few small things that need tweaking, and I have some suggestions for cosmetic changes.

src/mxnet/Handle.d
jens-mueller-sociomantic: So it's fine to have this output to `stderr` during testing? I find it a bit strange.

src/mxnet/Handle.d
jens-mueller-sociomantic: Yes. That should be a `,`. Good that you found this.

src/mxnet/Handle.d
joseph-wakeling-sociomantic: Yes, it's fine.  How about a message like:

`("Debug reporting works: {}\n", true)`

... ? ;-)

src/mxnet/Handle.d
jens-mueller-sociomantic: Makes sense. But I'd like to keep this one focused on refactoring.

src/mxnet/Handle.d
jens-mueller-sociomantic: I think it refers to all allocations (including GC allocations). I think `write` does not (heap) allocate. But I try to rephrase.

jens-mueller-sociomantic:
This change addresses the raised issues.

```diff
diff --git a/src/mxnet/Handle.d b/src/mxnet/Handle.d
index c5e271b..1f1a488 100644
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -98,8 +98,8 @@ static ~this ()
 {
     if (mxnet_handle_count != 0)
     {
-        stderrWritef("{} MXNet handle(s) in use at shutdown!\n",
-                     mxnet_handle_count);
+        debugReport("{} MXNet handle(s) in use at shutdown!\n",
+                    mxnet_handle_count);
     }
 }

@@ -378,27 +378,21 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
             --mxnet_handle_count;
             debug (MXNetHandle)
             {
-                // should avoid allocation, which (whether GC or not) might
-                // cause problems if this method is called from a destructor
-                stderrWritef("Freeing handle: {:x} ({} still allocated)\n",
-                             this.c_api_handle, handleCount());
+                debugReport("Freeing handle: {:x} ({} still allocated)\n",
+                            this.c_api_handle, handleCount());
             }

             if (handleCount() < 0)
             {
-                // we cannot throw (even an Error) from here, since this
-                // method may be called by the destructor
-                stderrWritef("Allocated handle count is negative!\n");
+                debugReport("Allocated handle count is negative!\n");
                 abort();
             }
         }

         if (FreeHandleFunction(this.c_api_handle) != 0)
         {
-            // we cannot throw (even an Error) from here, since this
-            // method may be called by the destructor
-            stderrWritef("Could not free MXNet handle using {}@{}:{}\n" ~
-                         identifier!(FreeHandleFunction), __FILE__, __LINE__);
+            debugReport("Could not free MXNet handle using {}@{}:{}\n",
+                        identifier!(FreeHandleFunction), __FILE__, __LINE__);
             abort();
         }

@@ -428,8 +422,8 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
                 // have been freed by the time the destructor is initiated.
                 // This debug check should help in tracking down any handles
                 // for which this has not been done.
-                stderrWritef("Non-null handle {:x} in MXNetHandle destructor!\n",
-                             this.c_api_handle);
+                debugReport("Non-null handle {:x} in MXNetHandle destructor!\n",
+                            this.c_api_handle);
             }
         }
         this.freeHandle();
@@ -441,10 +435,15 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)

     Writes a formatted string to standard error

-    The only allocations this function performs are the ones necessary for
-    string conversions of the arguments. If no allocations are performed this
-    function is safe to be executed inside a destructor or other places where
-    no allocation is supposed to happen.
+    The function itself does not allocate. But the string conversion method
+    `toString` for arguments of user-defined type might. As such these
+    allocations depend on the implementation of `toString`. In summary the
+    execution of this function performs no allocation (including GC
+    allocations) unless an argument cannot be formatted without.
+
+    As such the function is designed for outputting debug messages in
+    circumstances where GC allocations and throwing is not permitted. Most
+    notably this includes the execution of destructors.

     Params:
         fmt = format string to use
@@ -452,7 +451,7 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)

 *******************************************************************************/

-private void stderrWritef (Args...) (cstring fmt, Args args)
+private void debugReport (Args...) (cstring fmt, Args args)
 {
     sformat((cstring str) { write(STDERR_FILENO, str.ptr, str.length); },
             fmt, args);
@@ -460,7 +459,8 @@ private void stderrWritef (Args...) (cstring fmt, Args args)

 unittest
 {
-    testNoAlloc(stderrWritef("the number {}\n", 1));
+    testNoAlloc(debugReport("Debug reporting works: {} at {}:{}\n",
+                            true, __FILE__, __LINE__));
 }
```

joseph-wakeling-sociomantic (approved):
LGTM in principle, but a note: aren't there other error reports that use `Stderr.formatln`?  It might be worth updating them too.

Note that one thing to be careful of there is that the default formatting of pointers changes between `Stderr.formatln` and `sformat` (this is why the explicit `{:x}` is in place in the debug reports already covered by this patch).  So you might need to tweak things there in order to keep consistency of pointer formatting.

joseph-wakeling-sociomantic:
Might also be worth tweaking the 1st commit title and description a little to make clearer the motivation for the function (rather than just what it does).

jens-mueller-sociomantic:
Yes, there are two instances of `Stderr.formatln`. But as these are not executed during destruction I would keep them as is.

joseph-wakeling-sociomantic:
> Yes, there are two instances of `Stderr.formatln`. But as these are not executed during destruction I would keep them as is.

There's two reasons I'd like to switch them over: (i) consistency of meaning (i.e. it makes clear that all of these outputs are debug reports), and (ii) consistency of formatting (they all use the same fundamental mechanism, so no risk of discrepancy between them).

jens-mueller-sociomantic:
```diff
diff --git a/src/mxnet/Handle.d b/src/mxnet/Handle.d
index 1f1a488..3d1555d 100644
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -319,9 +319,9 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
         {
             debug (MXNetHandle)
             {
-                Stderr.formatln("Attempting to overwrite handle {} with itself "
+                debugReport("Attempting to overwrite handle {:x} with itself "
                                     ~ "at {}:{}",
-                                this.c_api_handle, file, line);
+                            this.c_api_handle, file, line);
             }
             return;
         }
@@ -335,8 +335,8 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
             ++mxnet_handle_count;
             debug (MXNetHandle)
             {
-                Stderr.formatln("New handle: {} ({} allocated in total)",
-                                this.c_api_handle, handleCount());
+                debugReport("New handle: {:x} ({} allocated in total)",
+                            this.c_api_handle, handleCount());
             }
         }
     }
```

src/mxnet/Handle.d
```diff
@@ -337,8 +335,8 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
             ++mxnet_handle_count;
             debug (MXNetHandle)
             {
-                Stderr.formatln("New handle: {} ({} allocated in total)",
-                                this.c_api_handle, handleCount());
+                debugReport("New handle: {:x} ({} allocated in total)",
```
joseph-wakeling-sociomantic: Missing `\n` (needed now you're switching to the new functionality).  Might be worth rolling that up under the hood of the new `debugReport` function?

src/mxnet/Handle.d
```diff
@@ -321,9 +319,9 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
         {
             debug (MXNetHandle)
             {
-                Stderr.formatln("Attempting to overwrite handle {} with itself "
+                debugReport("Attempting to overwrite handle {:x} with itself "
                                     ~ "at {}:{}",
```
joseph-wakeling-sociomantic: Missing `\n` (needed now you're switching to the new functionality).  Might be worth rolling that up under the hood of the new `debugReport` function?

joseph-wakeling-sociomantic (changes requested):
Perils of switching from `Stderr.formatln` to `write`: you're missing `\n` for the two new `debugReport` calls.

src/mxnet/Handle.d
jens-mueller-sociomantic: But how without allocating? Is there a simple way?

jens-mueller-sociomantic:
Rebased on `v0.x.x`

jens-mueller-sociomantic:
Adding the missing `\n`.

```diff
diff --git a/src/mxnet/Handle.d b/src/mxnet/Handle.d
index 9fbbace..74e7612 100644
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -321,7 +321,7 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
             debug (MXNetHandle)
             {
                 debugReport("Attempting to overwrite handle {:x} with itself "
-                                    ~ "at {}:{}",
+                                    ~ "at {}:{}\n",
                             this.c_api_handle, file, line);
             }
             return;
@@ -336,7 +336,7 @@ public class MXNetHandle (HandleType, alias FreeHandleFunction)
             ++mxnet_handle_count;
             debug (MXNetHandle)
             {
-                debugReport("New handle: {:x} ({} allocated in total)",
+                debugReport("New handle: {:x} ({} allocated in total)\n",
                             this.c_api_handle, handleCount());
             }
         }
```

joseph-wakeling-sociomantic (approved):
LGTM, feel free to take in at your convenience once the tests have passed.

PR #23, opened 2017-09-18
Fix documentation for right-hand side of operator *=

jens-mueller-sociomantic:
The currently implemented operator `*=` for n-dimensional arrays expects
the right-hand side to be a scalar. But the documentation specified it as
an n-dimensional array. This change fixes the documentation stating that
the right-hand is a scalar that is multiplied element-wise.


alaksiej-stankievicx-dunnhumby:
@jens-mueller-sociomantic, could you provide a kind of reference for bare mxnet to verify that it defines "*=" as newly described.

jens-mueller-sociomantic:
`Stdout(atomicSymbolInfo("_mul_scalar"));` outputs

```
_mul_scalar

2 parameters:
  * data : NDArray-or-Symbol
      - source input
  * scalar : float
      - scalar input
```

We use `_mul_scalar` to implement `*=`, similar to the [C++ bindings](https://github.com/apache/incubator-mxnet/blob/ec71cff6bc7aa02bc6d507ad458a2806d624a9e4/cpp-package/include/mxnet-cpp/ndarray.hpp#L157).

I can look for more. I couldn't find something suitable in the Python API. I could only find a [multiplication with broadcast](https://mxnet.incubator.apache.org/api/python/ndarray.html#mxnet.ndarray.multiply). Maybe they are moving away from the operations requiring float to string conversions (which gives another option for sociomantic-tsunami#22).

alaksiej-stankievicx-dunnhumby (approved):

jens-mueller-sociomantic:
@joseph-wakeling-sociomantic Any feedback from your side?

joseph-wakeling-sociomantic (approved):
LGTM.  Weird that the doc was like this :-\

PR #27, opened 2017-09-25
Upgrade to v3 of dlang dockerfile

joseph-wakeling-sociomantic:
This upgraded dockerfile should come with explicitly pinned versions of various packages (e.g. D compilers).  This should make for a more stable testing environment.

### DIFF
--- a/beaver.Dockerfile
+++ b/beaver.Dockerfile
@@ -1,3 +1,3 @@
-FROM sociomantictsunami/dlang:v2
+FROM sociomantictsunami/dlang:v3
 COPY docker/ /docker-tmp
 RUN /docker-tmp/build && rm -fr /docker-tmp


joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic you might want to confirm that I've correctly understood the impact of this upgrade ;-)

leandro-lucarella-sociomantic (approved):
Yeah, shocking and breaking changes!

leandro-lucarella-sociomantic:
(you might want to update beaver too if you didn't, it shouldn't be actually necessary unless you are using test docker images though)

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic already done, the PR was there less than 10 minutes after your release email ;-) https://github.com/sociomantic-tsunami/dmxnet/pull/26

Just for my understanding: can we now explicitly pin packages ourselves in our own dockerfile definitions?

leandro-lucarella-sociomantic:
Yes, you could always do that actually. But when you install your own packages you can chose the version, this is only useful for base images, where a user could do a `full-upgrade` and screw everything.

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic gotcha.  Thanks for the clarification :-)

PR #28, opened 2017-10-06
Update ocean from v3.3.3 to v3.4.3

jens-mueller-sociomantic:
* submodules/ocean v3.3.3(5a0f31e)...v3.4.3(d114d14) (36 commits)
  > Import core.stdc.stdarg in Layout[_tango]
  > Merge tag v3.3.4 into v3.4.x
  > Update to cachalot v3 image
  > Update to beaver v0.2.2
  > Merge tag v2.10.1 into v3.4.x
  (...)


joseph-wakeling-sociomantic:
Note, as a general rule, we shouldn't be updating the ocean submodule for a library (since it's important to avoid accidental breaking changes for downstreams, e.g. by relying on a more recent ocean than they make use of).  But we're in 0.x.x at the moment, so we can upgrade as is convenient for us ;-)

joseph-wakeling-sociomantic (approved):

PR #29, opened 2017-10-06
Fix deprecated use of `stop` since ocean v3.4.0

jens-mueller-sociomantic:
Since ocean v3.4.0 using `stop` is deprecated as it does not actually
stop the stop watch. This change replaces it with a call to `sec` which
is the recommended migration.


joseph-wakeling-sociomantic:
Would suggest a couple of tweaks to the commit title and message: put the full name (`StopWatch.stop`) in the commit title (there could be lots of `stop` functions/methods in library space!), and don't worry about `since ocean v3.4.0` in the title (it's sufficient to explain in the commit body as you have).

For English phrasing `Replace use of deprecated StopWatch.stop method` sounds better to me, but that's a small issue.

joseph-wakeling-sociomantic (approved):
LGTM, thanks!

PR #31, opened 2017-10-27
Add symbol representing a generalized dot product

jens-mueller-sociomantic:
This change adds a `Dot` symbol class representing a generalized dot
product of its inputs. This enables forming networks which need to
calculate dot products.


src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+
+        Params:
+            x = symbol whose last dimension's length must match `y` first
+                dimension's length
+            y = symbol whose first dimension's length must match the length of
+                the last dimension of `x`
+            transpose_x = transpose `x` prior to performing the dot; defaults
+                          to false
+            transpose_y = transpose `y` prior to performing the dot; defaults
+                          to false
```
joseph-wakeling-sociomantic: How do the dimension length restrictions change if `x` and or `y` are to be transposed?

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+        istring[2] keys;
+        keys[0] = "transpose_a";
+        keys[1] = "transpose_b";
+
+        Immut!(char)*[2] c_keys;
+        c_keys[0] = keys[0].ptr;
+        c_keys[1] = keys[1].ptr;
+
+        Const!(char)*[2] c_values;
+        const istring[] true_and_false = ["false", "true"];
```
joseph-wakeling-sociomantic: This seems a rather inefficient way for mxnet to serialize a true/false flag :-P

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+    {
+        istring[2] keys;
+        keys[0] = "transpose_a";
+        keys[1] = "transpose_b";
+
+        Immut!(char)*[2] c_keys;
+        c_keys[0] = keys[0].ptr;
+        c_keys[1] = keys[1].ptr;
+
+        Const!(char)*[2] c_values;
```
joseph-wakeling-sociomantic: This can be `Immut` too, I think ... ?

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+        args[1] = y.handle();
+
+        this.compose(args, null);
+    }
+
+    unittest
+    {
+        scope x = new Variable("x");
+        scope (exit) x.freeHandle();
+        scope y= new Variable("y");
```
joseph-wakeling-sociomantic: Minor style: space between `y` and `=`

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+        scope (exit) y.freeHandle();
+        scope dot = new Dot(x, y);
+        scope (exit) dot.freeHandle();
+        test!("!is")(dot.handle(), null);
```
joseph-wakeling-sociomantic: Minor note (more for the future than for this PR): this validates instantiation of the symbol, but not really its use/effect.

I wonder whether it might be a good idea to have a separate module of unittests that explore the combination of symbols and imperative code, just so we have some sanity-checking of basic expected behaviours.  It needn't be anything too complicated, just some minimal cases (e.g. for `Dot`, we might have one case for dot product of a pair of vectors, and one case for a couple of tiny `NdArray`s).

Something to think about for the future?

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,93 @@ public class LinearRegressionOutput : Symbol
...
+
+
+/*******************************************************************************
+
+    Dot symbol representing a generalized dot product
+
+    A sum is performed over the last dimension of the first operand and the
+    first dimension of the second operand. The reduction of the sum results in
+    an n-dimensional array whose dimensionality is
+    `x.shape.length - y.shape.length - 2`.
```
joseph-wakeling-sociomantic: I'm not sure I quite follow what is meant by this.  It's not clear to me what is meant by `The reduction of the sum`.

I don't mind taking as-is if there's a need to have the functionality in the library sooner rather than later, but it would be good to try to be a bit clearer with the description.  A simple example might help?

joseph-wakeling-sociomantic (changes requested):
Code wise seems fine bar one or two nitpicks.  The documentation seems like it could be made a bit clearer, though.

src/mxnet/Symbol.d
jens-mueller-sociomantic: How about `the reduction performed by the sum`? What I want to express is that the `sum` combines the two inputs along a dimension which is collapsed. I think of a better way of expressing this.

joseph-wakeling-sociomantic: Well TBH I'm not sure it's less opaque to me :-\  That's why I think an example might help.

Basically I don't follow from the description given why this process winds up with an `NdArray` with the dimensionality described.  FWIW I'm not sure I even follow how that can be true for simple 1D vectors ... ?

src/mxnet/Symbol.d
jens-mueller-sociomantic: I believe they reverse as well. If `x` is transposed its the length of the length of the *first* dimension of the original `x` must match the length length of the first dimension of `y`. I mean transposition reverses the `shape`. I'll guess I can make it clearer.

src/mxnet/Symbol.d
jens-mueller-sociomantic: The effect of this can only be seen when making passing over a complete network. What we can do is implement dot for n-dimensional arrays with `unittest`s which should show what it calculates.

joseph-wakeling-sociomantic:
A thought, since I guess you probably would like to expedite getting this into the library sooner rather than later ...

How about leaving the detailed doc for later and just documenting the class as:

```
Dot symbol representing a generalized dot product between two n-dimensional arrays
```

joseph-wakeling-sociomantic:
... and maybe the length constraints on the constructor parameters can be similarly left for later documentation.

jens-mueller-sociomantic:
Can do but I think the fixes we need are not too many to get it done now.

src/mxnet/Symbol.d
```diff
@@ -939,13 +939,22 @@ public class LinearRegressionOutput : Symbol
     Dot symbol representing a generalized dot product
 
     A sum is performed over the last dimension of the first operand and the
-    first dimension of the second operand. The reduction of the sum results in
-    an n-dimensional array whose dimensionality is
-    `x.shape.length - y.shape.length - 2`.
+    first dimension of the second operand. These dimensions have to be of same
+    length as they are collapsed by summing over the pair-wise products. The
+    result is an n-dimensional array whose shape is the shape of the input
+    n-dimensional arrays concatenated excluding the two dimensions used when
```
joseph-wakeling-sociomantic: Does this mean that the correct value of the original formula was `x.shape.length + y.shape.length - 2` ... ?

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,102 @@ public class LinearRegressionOutput : Symbol
...
+    dimensions are collapsed. The element at `[i_1, ..., i_m, j_1, ..., j_m]`
+    of the result is calculated by summing over the products
+
+        `x[i_1, ..., i_m, k] * y[k, j_1, ..., j_n]`
+
+    for `k` in `[0, ..., y.shape[0] - 1]`.
+
+*******************************************************************************/
+
+class Dot : Symbol
```
joseph-wakeling-sociomantic: Missing protection attribute

src/mxnet/Symbol.d
```diff
@@ -932,3 +932,102 @@ public class LinearRegressionOutput : Symbol
...
+    unittest
+    {
+        scope x = new Variable("x");
+        scope (exit) x.freeHandle();
+        scope y = new Variable("y");
+        scope (exit) y.freeHandle();
+        scope dot = new Dot(x, y);
+        scope (exit) dot.freeHandle();
+        test!("!is")(dot.handle(), null);
+    }
```
joseph-wakeling-sociomantic, line 1032: Why is the `unittest` block inline?  It means this test will run many more times than is necessary.

joseph-wakeling-sociomantic (commented):
The doc changes LGTM.  A few small tweaks and we can take this in.

src/mxnet/Symbol.d
jens-mueller-sociomantic: Oh, yes. I wrote `x.shape.length - y.shape.length - 2` which makes no sense.

src/mxnet/Symbol.d
jens-mueller-sociomantic: How can it run more than once? I thought this only happens for templated types.

joseph-wakeling-sociomantic: Maybe so.  It just feels weird having the unittests nested when they don't need to be.  At least with the templated types there's a reason for doing that, which is to make sure that the tests get run for all used instantiations.

joseph-wakeling-sociomantic: But since that seems to be normal in this module, I don't mind leaving it for a later decision.  Let's fix the protection attribute and we can get this in.

jens-mueller-sociomantic:
Squashed and the missing protection attribute.

```diff
diff --git a/src/mxnet/Symbol.d b/src/mxnet/Symbol.d
index 3b591ad..73450d1 100644
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -969,7 +969,7 @@ public class LinearRegressionOutput : Symbol

 *******************************************************************************/

-class Dot : Symbol
+public class Dot : Symbol
 {
     /***************************************************************************
```

joseph-wakeling-sociomantic (approved):
LGTM!  Let's get this in :-)

PR #35, opened 2017-11-29
Fix usage of SoftmaxOutput multi_output constructor parameter flag

joseph-wakeling-sociomantic:
This looks like a copy-paste error from back when the library was first created.

### DIFF
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -810,7 +810,7 @@ public class SoftmaxOutput : Symbol
         cstring ignore_label_str = toNoLossString(ignore_label, buf_ignore_label);
         buf_ignore_label[ignore_label_str.length] = '\0';
 
-        istring multi_output_str = use_ignore ? "true" : "false";
+        istring multi_output_str = multi_output ? "true" : "false";
 
         istring preserve_shape_str = preserve_shape ? "true" : "false";
 


joseph-wakeling-sociomantic:
As a separate note: is there any particular importance to the ordering of constructor parameters, or the ordering of key/value pairs?  There are some minor mismatches, which make no functional difference but make it a little bit more annoying to visually scan through everything.

joseph-wakeling-sociomantic:
Note, this fix also raises the question: is there any way for us to validate the settings associated with a symbol?  It would be good to be able to validate via unittests that these parameter settings are properly respected.

jens-mueller-sociomantic (approved):
Yes, seems like an obvious copy&paste error. Good that you found it.

jens-mueller-sociomantic:
> As a separate note: is there any particular importance to the ordering of constructor parameters, or the ordering of key/value pairs? There are some minor mismatches, which make no functional difference but make it a little bit more annoying to visually scan through everything.

The order is currently just the way MXNet outputs them (when generating the symbol documentation) I believe. You can reorder as you like. I think I didn't because it seems to be consistent across all MXNet bindings.

> Note, this fix also raises the question: is there any way for us to validate the settings associated with a symbol? It would be good to be able to validate via unittests that these parameter settings are properly respected.

You can query a symbol for its attributes (see https://github.com/sociomantic-tsunami/dmxnet/blob/v0.x.x/src/mxnet/c/c_api.d#L662, https://github.com/sociomantic-tsunami/dmxnet/blob/v0.x.x/src/mxnet/c/c_api.d#L637 and https://github.com/sociomantic-tsunami/dmxnet/blob/v0.x.x/src/mxnet/c/c_api.d#L654). It seems to provide the information you are asking for.
Whether the configured values are respected is another question. Because for this you have to use the symbol and verify that behaves as configured.

Issue #25, opened 2017-09-20
Generalize `toNoLossString` to handle arbitrary parameter types


joseph-wakeling-sociomantic:
`mxnet.Util.toNoLossString` was written solely with `float` variables in mind, but in practice there are multiple different types that might be of interest.

With this in mind, it could be a good idea to introduce a more general function for converting arbitrary numerical types to strings, without losing precision in the case of floating point.

As a suggestion, it might be a good idea to replace `mxnet.Util` with a new `mxnet.Parameters` module specifically dedicated to handling conversions to what the C API expects.  The new function would then have a form something like,

```D
public mstring parameterStringOf (T) (T value, mstring buf)
{
    static if (isFloatingPointType!(T))
    {
        return parameterStringOfFloat(value, buf);
    }
    else static if (isIntegerType!(T))
    {
        return parameterStringOfInteger(value, buf);
    }
    else
    {
        static assert(false, "parameterStringOf: unsupported type " ~ T.stringof);
    }
}
```

with individual `private` implementations on a parameter by parameter basis.

This should partially (but not completely) address issues raised in https://github.com/sociomantic-tsunami/dmxnet/pull/22.

jens-mueller-sociomantic:
I believe you only need the floating point type `float`, the integer type `uint` and the `bool` type. That should be all types of values that need to be converted to C strings.

joseph-wakeling-sociomantic:
Well, `NDArray` _nominally_ supports `float`, `double`, `int` and `ubyte` as element types, which suggests that (unless MXNet's own C API is broken), we need `parameterStringOf` to support all of these, at the very least just for the `_set_value` operation.

Note that supporting all floating-point types and all integer types inherently covers all of these, and we can easily add `bool` to the mix as well.

If e.g. `_set_value` doesn't support non-`float` types, that's a different issue, and one that we should handle separately: we shouldn't make `parameterStringOf` the gatekeeper.

joseph-wakeling-sociomantic:
Note, this shouldn't particularly be a hard problem, it's just a generalization of what is already done in `toNoLossString`, where we need a compile-time function or template in order to calculate the minimum buffer length depending on type.

jens-mueller-sociomantic:
> If e.g. `_set_value` doesn't support non-`float` types, that's a different issue, and one that we should handle separately: we shouldn't make `parameterStringOf` the gatekeeper.

[...]

PR #40, opened 2017-12-04
Replace separate SoftmaxOutput constructor parameters with Config struct 

joseph-wakeling-sociomantic:
This should make it much easier to customize one or two of the softmax config parameters, for example:

```D
SoftmaxOutput.Config config = { preserve_shape = true };
auto s = new SoftmaxOutput(input, label, config);
```

A usage example unittest has been added.

### DIFF
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -763,6 +763,109 @@ public enum SoftmaxOutputNormalization
 
 public class SoftmaxOutput : Symbol
 {
+    import ocean.core.Traits;
+
+    /***************************************************************************
+
+        Struct containing configuration options for the `SoftmaxOutput`
+        symbol
+
+        By default, all values will match the defaults used by the upstream
+        MXNet library.  This struct only needs to be used directly if one or
+        more of these parameters is to be set to a non-default value.
+
+    ***************************************************************************/
+
+    public struct Config
+    {
+        /***********************************************************************
+
+            Scale factor for scaling the gradient
+
+        ***********************************************************************/
+
+        public float grad_scale = 1;
+
+        /***********************************************************************
+
+            Label to ignore during the backward pass, if `use_ignore` is `true`
+
+        ***********************************************************************/
+
+        public float ignore_label = -1;
+
+        /***********************************************************************
+
+            If `true`, the softmax function will be computed along the second
+            axis, i.e. `axis == 1`
+
+        ***********************************************************************/
+
+        public bool multi_output = false;
+
+        /***********************************************************************
+
+            If `true`, the `ignore_label` field can be used to specify a label
+            to ignore during the backward pass
+
+        ***********************************************************************/
+
+        public bool use_ignore = false;
+
+        /***********************************************************************
+
+            If `true`, the softmax function will be computed along the last
+            axis, i.e. `axis == -1`
+
+        ***********************************************************************/
+
+        public bool preserve_shape = false;
+
+        /***********************************************************************
+
+            Normalization applied to the gradient
+
+        ***********************************************************************/
+
+        public SoftmaxOutputNormalization normalization =
+            SoftmaxOutputNormalization.batch;
+
+        /***********************************************************************
+
+            If `true`, the gradient will be multiplied elementwise by the
+            output gradient
+
+        ***********************************************************************/
+
+        public bool out_grad = false;
+    }
+
+    ///
+    unittest
+    {
+        // initialize a `Config` struct instance with only a couple of
+        // non-default values
+        Config config = {
+            ignore_label: 1,
+            use_ignore: true
+        };
+
+        // check values of custom fields
+        assert(config.ignore_label == 1.0f);
+        assert(config.ignore_label != Config.init.ignore_label);
+
+        assert(config.use_ignore);
+        assert(config.use_ignore != Config.init.use_ignore);
+
+        // check values of non-customized fields
+        assert(config.grad_scale == Config.init.grad_scale);
+        assert(config.multi_output == Config.init.multi_output);
+        assert(config.preserve_shape == Config.init.preserve_shape);
+        assert(config.normalization == Config.init.normalization);
+        assert(config.out_grad == Config.init.out_grad);
+    }
+
+
     /***************************************************************************
 
         Constructs `SoftmaxOutput` symbol
@@ -770,29 +873,15 @@ public class SoftmaxOutput : Symbol
         Params:
             input = input symbol to apply softmax to
             label = ground truth to compare against the output of softmax
-            normalization = normalization applied to the gradient; defaults to
-                            batch
-            grad_scale = scale factor for scaling the gradient; defaults to 1
-            use_ignore = use ignore_label; defaults to false
-            ignore_label = all labels with this label will be ignored during the
-                           backward pass; defaults to -1
-            multi_output = softmax applied to axis 1, if set to true; defaults
-                           to false
-            preserve_shape = softmax will applied on the last axis, if true;
-                             defaults to false
-            out_grad = apply weighting to output gradient
+            config = optional instance of Config struct containing extra
+                     configuration parameters (if not specified, all will
+                     be set to their default values)
 
     ***************************************************************************/
 
     public this (Symbol input,
                  Symbol label,
-                 SoftmaxOutputNormalization normalization = SoftmaxOutputNormalization.batch,
-                 float grad_scale = 1,
-                 bool use_ignore = false,
-                 float ignore_label = -1,
-                 bool multi_output = false,
-                 bool preserve_shape = false,
-                 bool out_grad = false)
+                 Config config = Config.init)
     in
     {
         assert(input !is null);
@@ -800,44 +889,45 @@ public class SoftmaxOutput : Symbol
     }
     body
     {
-        char[16] buf_grad_scale = void;
-        cstring grad_scale_str = toNoLossString(grad_scale, buf_grad_scale);
-        buf_grad_scale[grad_scale_str.length] = '\0';
-
-        istring use_ignore_str = use_ignore ? "true" : "false";
-
-        char[16] buf_ignore_label = void;
-        cstring ignore_label_str = toNoLossString(ignore_label, buf_ignore_label);
-        buf_ignore_label[ignore_label_str.length] = '\0';
-
-        istring multi_output_str = multi_output ? "true" : "false";
-
-        istring preserve_shape_str = preserve_shape ? "true" : "false";
-
-        istring out_grad_str = out_grad ? "true" : "false";
-
-        istring[7] keys;
-        keys[0] = "grad_scale";
-        keys[1] = "ignore_label";
-        keys[2] = "multi_output";
-        keys[3] = "use_ignore";
-        keys[4] = "preserve_shape";
-        keys[5] = "normalization";
-        keys[6] = "out_grad";
-
-        Immut!(char)*[7] c_keys;
-        foreach (i, ref key; keys) c_keys[i] = key.ptr;
-
         const istring[] softmax_normalizations = ["batch", "null", "valid"];
 
-        Const!(char)*[7] c_values;
-        c_values[0] = grad_scale_str.ptr;
-        c_values[1] = ignore_label_str.ptr;
-        c_values[2] = multi_output_str.ptr;
-        c_values[3] = use_ignore_str.ptr;
-        c_values[4] = preserve_shape_str.ptr;
-        c_values[5] = softmax_normalizations[normalization].ptr;
-        c_values[6] = out_grad_str.ptr;
+        alias typeof(Config.tupleof) ConfigTuple;
+
+        istring[ConfigTuple.length] keys;
+        Const!(char)*[ConfigTuple.length] c_keys;
+        Const!(char)*[ConfigTuple.length] c_values;
+
+        foreach (i, T; ConfigTuple)
+        {
+            keys[i] = FieldName!(i, Config);
+            c_keys[i] = keys[i].ptr;
+
+            auto field_value = config.tupleof[i];
+
+            static if (isFloatingPointType!(T))
+            {
+                char[16] float_buf = void;
+                cstring float_string = toNoLossString(field_value, float_buf);
+                float_buf[float_string.length] = '\0';
+                c_values[i] = float_string.ptr;
+
+            }
+            else static if (is(T == bool))
+            {
+                istring bool_string = field_value ? "true" : "false";
+                c_values[i] = bool_string.ptr;
+            }
+            else static if (is(T == SoftmaxOutputNormalization))
+            {
+                istring norm_string = softmax_normalizations[field_value];
+                c_values[i] = norm_string.ptr;
+            }
+            else
+            {
+                static assert("Unsupported SoftmaxOutput config type "
+                              ~ T.stringof);
+            }
+        }
 
         super("SoftmaxOutput", c_keys, c_values);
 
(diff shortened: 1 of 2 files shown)


joseph-wakeling-sociomantic:
This is missing release notes, but I'd like to get some review of the code approach first to make sure it's acceptable.

joseph-wakeling-sociomantic:
As a first submission, I've broken the changes up into transitional patches so that it's possible to validate the logic (and equivalence) of the old and new approaches.

BTW, note that this would be even simpler if we could address https://github.com/sociomantic-tsunami/dmxnet/issues/25 and roll up some of the special-casing under the hood of a general `toParameterString`.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+        public bool preserve_shape = false;
+
+        /***********************************************************************
+
+            Normalization applied to the gradient (defaults to `batch`)
+
+        ***********************************************************************/
+
+        public SoftmaxOutputNormalization normalization =
+            SoftmaxOutputNormalization.batch;
```
joseph-wakeling-sociomantic, line 831: @jens-mueller-sociomantic it's probably best to leave this as is for this PR, but it would probably be simpler to rewrite this parameter to just use `istring` values.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+
+            Scale factor for scaling the gradient (defaults to 1)
+
+        ***********************************************************************/
+
+        public float grad_scale = 1;
+
+        /***********************************************************************
+
+            Label to ignore during backwards propagation, if `use_ignore` is
```
jens-mueller-sociomantic: I would keep the old way of saying `backward pass` instead of `backwards propagation`.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+            true (defaults to -1)
+
+        ***********************************************************************/
+
+        public float ignore_label = -1;
+
+        /***********************************************************************
+
+            If `true`, the softmax function will be computed along the second
+            axis, i.e. `axis == 1` (defaults to `false`)
```
jens-mueller-sociomantic: I think we never used the term `axis` so far. But should be clear, I'll guess.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+
+        ***********************************************************************/
+
+        public bool multi_output = false;
+
+        /***********************************************************************
+
+            If `true`, the `ignore_label` field can be used to specify a label
+            to ignore during backwards propagation (defaults to `false`)
```
jens-mueller-sociomantic: see above regarding `backwards propagation`.

joseph-wakeling-sociomantic:
Updated with release notes.  These patches can all be squashed together when review is complete.

src/mxnet/Symbol.d
```diff
@@ -860,6 +915,29 @@ public class SoftmaxOutput : Symbol
...
+    scope label = new Variable("y");
+    scope (exit) label.freeHandle();
+
+    // create SoftmaxOutput symbol using default configuration
+    scope softmax_default = new SoftmaxOutput(input, label);
+    scope (exit) softmax_default.freeHandle();
+
+    // create SoftmaxOutput symbol with custom configuration
+    SoftmaxOutput.Config config = {
+        grad_scale: 1.5,
```
jens-mueller-sociomantic, line 968: I would suggest here to explicitly write `1.5f`.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+        {
+            keys[i] = FieldName!(i, Config);
+            c_keys[i] = keys[i].ptr;
 
-        Immut!(char)*[7] c_keys;
-        foreach (i, ref key; keys) c_keys[i] = key.ptr;
+            auto field_value = config.tupleof[i];
 
-        const istring[] softmax_normalizations = ["batch", "null", "valid"];
+            static if (isFloatingPointType!(T))
```
jens-mueller-sociomantic: This series of `static if` looks useful beyond this usage here. Can you turn it into a function that be reused elsewhere?

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
-        keys[1] = "ignore_label";
-        keys[2] = "multi_output";
-        keys[3] = "use_ignore";
-        keys[4] = "preserve_shape";
-        keys[5] = "normalization";
-        keys[6] = "out_grad";
+        foreach (i, T; ConfigTuple)
+        {
+            keys[i] = FieldName!(i, Config);
+            c_keys[i] = keys[i].ptr;
```
jens-mueller-sociomantic: I'll guess `c_keys[i] = FieldName!(i, Config).ptr` won't compile?

jens-mueller-sociomantic (commented):
Looks good to me. I only have some suggestions.

src/mxnet/Symbol.d
joseph-wakeling-sociomantic: Yes, but I don't want to make this change conditional on that.  Right now it's an implementation detail.  We can rework using a more extensive `toParameterString` later.

jens-mueller-sociomantic: I only point it out because it makes the current code harder to follow compared to the old approach.

joseph-wakeling-sociomantic:
Tweaked a bunch of things in response to feedback:

```diff
diff --git a/src/mxnet/Symbol.d b/src/mxnet/Symbol.d
index 89d6157..b5f76f4 100644
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -769,7 +769,7 @@ public class SoftmaxOutput : Symbol
     {
         /***********************************************************************
 
-            Scale factor for scaling the gradient (defaults to 1)
+            Scale factor for scaling the gradient
 
         ***********************************************************************/
 
@@ -777,8 +777,8 @@ public class SoftmaxOutput : Symbol
 
         /***********************************************************************
 
-            Label to ignore during backwards propagation, if `use_ignore` is
-            true (defaults to -1)
+            Label to ignore during the backwards pass, if `use_ignore` is
+            `true`
 
         ***********************************************************************/
 
@@ -787,7 +787,7 @@ public class SoftmaxOutput : Symbol
         /***********************************************************************
 
             If `true`, the softmax function will be computed along the second
-            axis, i.e. `axis == 1` (defaults to `false`)
+            axis, i.e. `axis == 1`
 
         ***********************************************************************/
 
@@ -796,7 +796,7 @@ public class SoftmaxOutput : Symbol
         /***********************************************************************
 
             If `true`, the `ignore_label` field can be used to specify a label
-            to ignore during backwards propagation (defaults to `false`)
+            to ignore during the backwards pass
 
         ***********************************************************************/
 
@@ -805,7 +805,7 @@ public class SoftmaxOutput : Symbol
         /***********************************************************************
 
             If `true`, the softmax function will be computed along the last
-            axis, i.e. `axis == -1` (defaults to `false`)
+            axis, i.e. `axis == -1`
 
         ***********************************************************************/
 
@@ -813,7 +813,7 @@ public class SoftmaxOutput : Symbol
 
         /***********************************************************************
 
-            Normalization applied to the gradient (defaults to `batch`)
+            Normalization applied to the gradient
 
         ***********************************************************************/
 
@@ -858,14 +858,12 @@ public class SoftmaxOutput : Symbol
 
         alias typeof(Config.tupleof) ConfigTuple;
 
-        istring[ConfigTuple.length] keys;
         Const!(char)*[ConfigTuple.length] c_keys;
         Const!(char)*[ConfigTuple.length] c_values;
 
         foreach (i, T; ConfigTuple)
         {
-            keys[i] = FieldName!(i, Config);
-            c_keys[i] = keys[i].ptr;
+            c_keys[i] = FieldName!(i, Config).ptr;
 
             auto field_value = config.tupleof[i];
```

src/mxnet/Symbol.d
joseph-wakeling-sociomantic: I assumed it was unwise, but it seems to work.

joseph-wakeling-sociomantic:
On reflection, I've reinstated the `keys` array.  It seems saner/safer to not make assumptions that directly using `FieldName!(...).ptr` is valid.

```diff
diff --git a/src/mxnet/Symbol.d b/src/mxnet/Symbol.d
index b5f76f4..897740a 100644
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -858,12 +858,14 @@ public class SoftmaxOutput : Symbol
 
         alias typeof(Config.tupleof) ConfigTuple;
 
+        istring[ConfigTuple.length] keys;
         Const!(char)*[ConfigTuple.length] c_keys;
         Const!(char)*[ConfigTuple.length] c_values;
 
         foreach (i, T; ConfigTuple)
         {
-            c_keys[i] = FieldName!(i, Config).ptr;
+            keys[i] = FieldName!(i, Config);
+            c_keys[i] = keys[i].ptr;
 
             auto field_value = config.tupleof[i];
```

src/mxnet/Symbol.d
joseph-wakeling-sociomantic: Is there a particular reason?  It doesn't make any difference, and in general I'd resist qualifying numbers in this way unless it's strictly _necessary_.  Otherwise it just risks creating annoyances in future (if e.g. the floating-point type used changes).

jens-mueller-sociomantic:
@joseph-wakeling-sociomantic Can you squash before I review?

joseph-wakeling-sociomantic:
I would like to leave it un-squashed for review at first, so that you can track the sanity of the incremental changes and the validation that behaviour has not changed.  Will squash once that's confirmed.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+
+            Scale factor for scaling the gradient
+
+        ***********************************************************************/
+
+        public float grad_scale = 1;
+
+        /***********************************************************************
+
+            Label to ignore during the backwards pass, if `use_ignore` is
```
jens-mueller-sociomantic: We usually say just `backward` (without the `s`) pass. I don't know whether there is a grammatical difference.

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+            axis, i.e. `axis == 1`
+
+        ***********************************************************************/
+
+        public bool multi_output = false;
+
+        /***********************************************************************
+
+            If `true`, the `ignore_label` field can be used to specify a label
+            to ignore during the backwards pass
```
jens-mueller-sociomantic: ditto

src/mxnet/Symbol.d
```diff
@@ -763,81 +763,136 @@ public enum SoftmaxOutputNormalization
...
+            c_keys[i] = keys[i].ptr;
 
-        Immut!(char)*[7] c_keys;
-        foreach (i, ref key; keys) c_keys[i] = key.ptr;
+            auto field_value = config.tupleof[i];
 
-        const istring[] softmax_normalizations = ["batch", "null", "valid"];
+            static if (isFloatingPointType!(T))
+            {
+                char[16] float_buf;
```
jens-mueller-sociomantic: There used to be a `= void` here but I don't know whether this is necessary to explicitly skip initialization of the buffer. Maybe you also want to initialize this memory here?

src/mxnet/Symbol.d
jens-mueller-sociomantic: When reading this line, I would now assume that double precision is used. But in fact only single precision is because this is (silently) converted to a `float` value.

jens-mueller-sociomantic (commented):
Some minor suggestions. Nothing major.

src/mxnet/Symbol.d
joseph-wakeling-sociomantic: Well, it's a floating point value.  There's no harm in specifying it in default precision, knowing very well that it may be stored in something smaller.

Switch the situation around: suppose that the underlying implementation changes to `double`, now we have to (remember to) rewrite the call site as well?

joseph-wakeling-sociomantic: Basically, the suggestion here is that we avoid unnecessary use of qualifiers for numerical constants, because it creates an extra maintenance burden of things to rewrite _in the case that the type changes_, for not a lot of benefit in terms of readability.

The only situation where we should be worrying about that is where it actually _matters_ to specifically have a `float` (e.g. because it's necessary to match exact test results).

joseph-wakeling-sociomantic:
Updated with some requested tweaks:

```diff
diff --git a/src/mxnet/Symbol.d b/src/mxnet/Symbol.d
index 897740a..ad6cb5c 100644
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -777,8 +777,7 @@ public class SoftmaxOutput : Symbol
 
         /***********************************************************************
 
-            Label to ignore during the backwards pass, if `use_ignore` is
-            `true`
+            Label to ignore during the backward pass, if `use_ignore` is `true`
 
         ***********************************************************************/
 
@@ -796,7 +795,7 @@ public class SoftmaxOutput : Symbol
         /***********************************************************************
 
             If `true`, the `ignore_label` field can be used to specify a label
-            to ignore during the backwards pass
+            to ignore during the backward pass
 
         ***********************************************************************/
 
@@ -871,7 +870,7 @@ public class SoftmaxOutput : Symbol
 
             static if (isFloatingPointType!(T))
             {
-                char[16] float_buf;
+                char[16] float_buf = void;
                 cstring float_string = toNoLossString(field_value, float_buf);
                 float_buf[float_string.length] = '\0';
                 c_values[i] = float_string.ptr;
```

src/mxnet/Symbol.d
jens-mueller-sociomantic: Note though that sometimes the value cannot be converted properly. This is rare but it can happen. But here nothing should go wrong. I tend to lean more toward having it explicitly. In D a `double` is silently converted to a `float` which looks odd to me. But I understand your reasoning.

jens-mueller-sociomantic:
LGTM. Can be squashed.

joseph-wakeling-sociomantic:
Squashed.

jens-mueller-sociomantic (approved):

joseph-wakeling-sociomantic:
Whoops: forgot to document the `Config` struct itself.  Will fix.

joseph-wakeling-sociomantic:
Updated with docs (and a usage example unittest) for the `Config` struct:

```diff
diff --git a/src/mxnet/Symbol.d b/src/mxnet/Symbol.d
index ad6cb5c..039bb46 100644
--- a/src/mxnet/Symbol.d
+++ b/src/mxnet/Symbol.d
@@ -765,6 +765,17 @@ public class SoftmaxOutput : Symbol
 {
     import ocean.core.Traits;
 
+    /***************************************************************************
+
+        Struct containing configuration options for the `SoftmaxOutput`
+        symbol
+
+        By default, all values will match the defaults used by the upstream
+        MXNet library.  This struct only needs to be used directly if one or
+        more of these parameters is to be set to a non-default value.
+
+    ***************************************************************************/
+
     public struct Config
     {
         /***********************************************************************
@@ -829,6 +840,31 @@ public class SoftmaxOutput : Symbol
         public bool out_grad = false;
     }
 
+    ///
+    unittest
+    {
+        // initialize a `Config` struct instance with only a couple of
+        // non-default values
+        Config config = {
+            ignore_label: 1,
+            use_ignore: true
+        };
+
+        // check values of custom fields
+        assert(config.ignore_label == 1.0f);
+        assert(config.ignore_label != Config.init.ignore_label);
+
+        assert(config.use_ignore);
+        assert(config.use_ignore != Config.init.use_ignore);
+
+        // check values of non-customized fields
+        assert(config.grad_scale == Config.init.grad_scale);
+        assert(config.multi_output == Config.init.multi_output);
+        assert(config.preserve_shape == Config.init.preserve_shape);
+        assert(config.normalization == Config.init.normalization);
+        assert(config.out_grad == Config.init.out_grad);
+    }
+
 
     /***************************************************************************
```

jens-mueller-sociomantic (approved):

PR #41, opened 2017-12-04
Clarify documentation for executor `outputs` method

jens-mueller-sociomantic:
This change extends the documentation of the `outputs` method for the
`Executor`. It adds that the outputted n-dimensional arrays share the
underlying data over multiple calls. Hence, there is usually no need to
call the `outputs` method more than once. It also adds that any
underlying handle of passed n-dimensional arrays is replaced, i.e.,
their handles are freed.
Also the `Returns` section is removed as this method does not have a
return parameter and in the `Throws` section the type of the thrown
exception is corrected to `MXNetExecutorException` and a typo regarding
the ordering of words is fixed.
The `unittest` of the method is updated to verify the extended
documentation.


joseph-wakeling-sociomantic:
@jens-mueller-sociomantic this PR looks like it has more patches in than I'd expect.  Are you sure that it's set up right?

src/mxnet/Executor.d
```diff
@@ -396,19 +396,26 @@ public class Executor (T)
...
+        Sets the outputs generated by the last forward pass of this executor
 
-        Params:
-            outputs = array to store the outputs whose elements will be
-                      overwritten; the length of outputs must match the number
-                      of the calculated outputs and each element must be
-                      non-null
+        The underlying data of the returned outputs is the same for multiple
+        calls of this method. Only fresh handles are used but they refer to the
+        same memory. When performing a forward pass the underlying data is
```
joseph-wakeling-sociomantic: I'm not sure I follow this phrase: `Only fresh handles are used but they refer to the same memory.`  Can you explain in more detail?

jens-mueller-sociomantic:
I forgot to rebase. Should be good now.

src/mxnet/Executor.d
jens-mueller-sociomantic: The method will always return n-dimensional arrays with new handles. But these handles reference the same memory. In essence, calling this method more than once gives you new n-dimensional arrays that reference the same memory.

joseph-wakeling-sociomantic: The method doesn't return anything.  Can you be more precise?

It sounds like what you are saying is that the underlying handles of the provided `NDArray` instances will be overwritten with handles that belong to the `Executor`, but which point into the same buffer of memory as the original handles.  Is that correct?

jens-mueller-sociomantic: Regarding the first part yes, regarding the second I think no. The original handles might be even `null`. But the point is that the output handles will always refer to same memory. Though the handles might be different. I don't know why they won't return the same handles actually. Probably because then I would need to remember what they returned. So it's simpler just to return a new C++ NDArray that references the same memory.

src/mxnet/Executor.d
```diff
@@ -396,19 +396,26 @@ public class Executor (T)
 
     /***************************************************************************
 
-        Returns the outputs generated by the last forward pass of this executor
+        Sets the outputs generated by the last forward pass of this executor
```
joseph-wakeling-sociomantic: It kind of feels like what we need to say here is something like:

```
    Sets the underlying handles of the provided `NDArray` instances to
    point to the output data generated by the executor's forward passes.
    As a result, after each forward pass, the results can be accessed via
    these `NDArray` instances.
```

joseph-wakeling-sociomantic: Then, the follow-up things to emphasize are:

* if the passed `NDArray` instances already have associated handles, those will be freed

  * note that we could actually rework things so that one is obliged to provide `NDArray` instances without a pre-existing handle

* if one passes multiple different `NDArray` instances then this creates a risky situation because you have multiple handles to the same data.  (I presume this isn't risky from a MXNet point of view, because presumably the executor's output buffers are its own; but from the wrapper's point of view duplicate handles are a problem.)

jens-mueller-sociomantic: Sounds good. I'd like to add explicitly though that any handle of a passed in `NDArray` instance is replaced, meaning a user cannot access these anymore. And also the case when you call it more than once. Because the data is shared which might be surprizing. A user might expect that is copied.

jens-mueller-sociomantic: I'm not sure that having two distinct handles referencing the same memory is risky. I believe we nowhere assume or state that two handles are not allowed to share the same memory.

joseph-wakeling-sociomantic: The risk is on the D wrapper side: that freeing the handle of one `NDArray` instance frees it for another.  No ... ?

joseph-wakeling-sociomantic:
> I'd like to add explicitly though that any handle of a passed in NDArray instance is replaced, meaning a user cannot access these anymore.

Is there any reason why we shouldn't just outright ban calling `outputs` with an `NDArray` that has a pre-existing handle?

joseph-wakeling-sociomantic:
Note that a ban on doing so doesn't prevent the user from recycling/reusing `NDArray` instances.  It just puts the onus on the user to manually free the handles before calling `outputs`.

That's probably saner/safer because then the user has to make a clear decision about whether or not it's safe to do so.  It won't just happen magically under the hood.

jens-mueller-sociomantic:
> Is there any reason why we shouldn't just outright ban calling `outputs` with an `NDArray` that has a pre-existing handle?

I don't see any reason. Seems like a good change for another PR. The documentation can be updated then as needed.

src/mxnet/Executor.d
jens-mueller-sociomantic: No. That should not be the case. They are **different** handles. We just have to handles/references to the same memory and MXNet will free that memory when there are no references anymore I'll guess.

joseph-wakeling-sociomantic: Different handle instances referencing the same memory, you mean?  So if we do two different calls to `outputs` with two different arrays of `NDArray` instances, the resulting handle pointers will be different?

joseph-wakeling-sociomantic:
> Seems like a good change for another PR.

Might be worth just implementing the ban and updating the doc only once.  It simplifies what we need to write here, after all.

jens-mueller-sociomantic:
> Might be worth just implementing the ban and updating the doc only once. It simplifies what we need to write here, after all.

Doesn't it only affect one sentence?

jens-mueller-sociomantic:
This PR is about extending the documentation of the current implementation. I prefer having no change of semantics here.

src/mxnet/Executor.d
jens-mueller-sociomantic: Yes. Have you look at the updated `unittest`?

src/mxnet/Executor.d
```diff
@@ -457,6 +464,18 @@ public class Executor (T)
...
+        scope (exit) output2.freeHandle();
+        scope outputs2 = [output2];
+        executor.outputs(outputs2);
+
+        // different instances
+        test!("!is")(output, output2);
+        // with different handles
+        test!("!is")(output.handle, output2.handle);
+        // but sharing the same data
+        test!("is")(output.data, output2.data);
```
joseph-wakeling-sociomantic, line 479: It might be a good idea to explicitly free the handle of `output2` and validate that `output` is still valid.

src/mxnet/Executor.d
```diff
@@ -457,6 +464,18 @@ public class Executor (T)
         scope outputs = [output];
         executor.outputs(outputs);
         test!("==")(output.data, input.data);
+
+        scope output2 = new NDArray!(float)(null);
+        scope (exit) output2.freeHandle();
+        scope outputs2 = [output2];
+        executor.outputs(outputs2);
```
joseph-wakeling-sociomantic: Minor note: I wouldn't bother with this `scope outputs` business in a unittest, just call:

```D
executor.outputs([outputs2]);
```

If you really want to avoid the array allocation, it would be helpful to be more explicit in the difference in variable names, e.g.

```D
scope output2_arr = [output2];
```

src/mxnet/Executor.d
joseph-wakeling-sociomantic: The unittest doesn't (yet) validate that freeing one of the `NDArray` handles doesn't affect the validity of the other (and the data it points to).

joseph-wakeling-sociomantic:
> This PR is about extending the documentation of the current implementation. I prefer having no change of semantics here.

Fair enough.

src/mxnet/Executor.d
jens-mueller-sociomantic: Yes, sounds useful.

jens-mueller-sociomantic:
Squashed.

joseph-wakeling-sociomantic:
The opening phrase of the method's documentation feels ambiguous:

```
Sets the outputs generated by the last forward pass of this executor
```

... because it doesn't do this.  It sets the handles of the `NDArrays` to point to the memory locations storing the output data.  This is an important distinction.

jens-mueller-sociomantic:
Rebased.

src/mxnet/Executor.d
```diff
@@ -396,7 +396,7 @@ public class Executor (T)
 
     /***************************************************************************
 
-        Sets the outputs generated by the last forward pass of this executor
+        Updates `outputs` to refer to the output data of this executor
```
joseph-wakeling-sociomantic: At this point in the documentation we don't know what `outputs` is.  Probably better to describe in a more general sense, which we can then understand more specifically later in terms of the `Params:` doc.

src/mxnet/Executor.d
```diff
@@ -396,19 +396,28 @@ public class Executor (T)
...
-                      of the calculated outputs and each element must be
-                      non-null
+        The underlying handles of the provided `NDArray` instances are set to
+        point to the output data generated by the executor's forward passes.
+        As a result, after each forward pass, the results can be accessed via
+        these `NDArray` instances.
 
-        Returns:
-            the outputs generated by the last forward pass
+        When setting the handles any handle is freed (for that reason
```
joseph-wakeling-sociomantic:
```
Any pre-existing handles in the `NDArray` instances will be freed
```

and, `for this reason` ;-)

src/mxnet/Executor.d
```diff
@@ -396,19 +396,28 @@ public class Executor (T)
...
+        `NDArray`s instances with `null` handles are usually passed in). Each
+        time this method is called the handles of the passed instances are set.
+        When calling this method more than once with different instances you
+        end up with different `NDArray` instances with different handles but
+        referencing the same elements.
+
+        Params:
+            outputs = array whose elements will be adjusted to hold the outputs
```
joseph-wakeling-sociomantic: It is pretty crazy that we shadow _the name of the method_ with its own parameter name!

src/mxnet/Executor.d
```diff
@@ -396,19 +396,28 @@ public class Executor (T)
...
+        these `NDArray` instances.
 
-        Returns:
-            the outputs generated by the last forward pass
+        When setting the handles any handle is freed (for that reason
+        `NDArray`s instances with `null` handles are usually passed in). Each
+        time this method is called the handles of the passed instances are set.
+        When calling this method more than once with different instances you
+        end up with different `NDArray` instances with different handles but
+        referencing the same elements.
```
joseph-wakeling-sociomantic: A minor point of clarification.

Suppose that I call `outputs` twice in succession with the same array of `NDArrays`.

Do the handles get updated the second time?  Or will they remain the same, given that they already point to the executor's output data?

src/mxnet/Executor.d
```diff
@@ -396,19 +396,28 @@ public class Executor (T)
...
-                      non-null
+        The underlying handles of the provided `NDArray` instances are set to
+        point to the output data generated by the executor's forward passes.
+        As a result, after each forward pass, the results can be accessed via
+        these `NDArray` instances.
 
-        Returns:
-            the outputs generated by the last forward pass
+        When setting the handles any handle is freed (for that reason
+        `NDArray`s instances with `null` handles are usually passed in). Each
```
joseph-wakeling-sociomantic: Surely `it is recommended to pass in only instances with `null` handles` would be better phrasing (we have no control over what is _usual_).

src/mxnet/Executor.d
jens-mueller-sociomantic: We can rename it but not in this PR. I never ran into an issue with this. Probably `dmd` warnings as errors saved me.

src/mxnet/Executor.d
jens-mueller-sociomantic: The handles will be changed.

src/mxnet/Executor.d
joseph-wakeling-sociomantic: In any case I think we can probably clarify the documentation more.  Maybe something like

```
array of `NDArray` instances (must be exactly one per output variable calculated by
the executor) whose handles will be updated in-place to point to the memory buffers
into which the executor writes the values of its output variables.
```

My phrasing is a bit rough there, but you get the idea ... ?

src/mxnet/Executor.d
```diff
@@ -396,19 +396,28 @@ public class Executor (T)
...
+
+        Params:
+            outputs = array whose elements will be adjusted to hold the outputs
+                      (each handle is replaced by a fresh handle referring an
+                      output); the length of outputs must match the number of
+                      the calculated outputs and each element must be non-null
 
         Throws:
-            `MXNetException` if `outputs.length` is equal not to the number of
+            `MXNetException` if `outputs.length` is not equal to the number of
```
joseph-wakeling-sociomantic: Should be `MXNetExecutorException`, no ... ?

jens-mueller-sociomantic: Seems so.

jens-mueller-sociomantic:
Squashed.

jens-mueller-sociomantic:
What issues are left with respect to the changed documentation?

src/mxnet/Executor.d
```diff
@@ -396,20 +396,29 @@ public class Executor (T)
 
     /***************************************************************************
 
-        Returns the outputs generated by the last forward pass of this executor
+        Sets n-dimensional arrays to refer to the output data of this executor
```
joseph-wakeling-sociomantic: I'm trying to think of a really good way to phrase this.

How about:

```
Makes output data of the executor available via the provided `NDArray`
instances
```

... ?

src/mxnet/Executor.d
```diff
@@ -396,20 +396,29 @@ public class Executor (T)
...
+        these `NDArray` instances.
 
-        Returns:
-            the outputs generated by the last forward pass
+        Any pre-existing handles in the `NDArray` instances will be freed (for
+        this reason it is recommended to pass in only instances with `null`
+        handles). Each time this method is called the handles of the passed
+        instances are set.  When calling this method more than once with
+        different instances you end up with different `NDArray` instances with
+        different handles but referencing the same elements.
```
joseph-wakeling-sociomantic: `same underlying elements` might be a nice qualification.

joseph-wakeling-sociomantic:
I have a couple of small suggestions for tweaks, but otherwise this LGTM.

jens-mueller-sociomantic:
Added the suggested changes.

```diff
diff --git a/src/mxnet/Executor.d b/src/mxnet/Executor.d
index a56de80..2bd8a06 100644
--- a/src/mxnet/Executor.d
+++ b/src/mxnet/Executor.d
@@ -396,7 +396,8 @@ public class Executor (T)

     /***************************************************************************

-        Sets n-dimensional arrays to refer to the output data of this executor
+        Makes output data of the executor available via the provided `NDArray`
+        instances

         The underlying handles of the provided `NDArray` instances are set to
         point to the output data generated by the executor's forward passes.
@@ -408,7 +409,7 @@ public class Executor (T)
         handles). Each time this method is called the handles of the passed
         instances are set.  When calling this method more than once with
         different instances you end up with different `NDArray` instances with
-        different handles but referencing the same elements.
+        different handles but referencing the same underlying elements.

         Params:
             outputs = array of `NDArray` instances (must be exactly one per
```

joseph-wakeling-sociomantic (dismissed):

jens-mueller-sociomantic:
Rebased.

joseph-wakeling-sociomantic:
One super-small tweak to commit message (should have noticed before): could you replace `i.e., their handles are freed` with `with any pre-existing handles being freed`?

It seems to make sense to emphasize this, because otherwise the phrase could be ambiguous about exactly what happens.

jens-mueller-sociomantic:
Updated commit message.

joseph-wakeling-sociomantic (approved):

PR #47, opened 2018-01-03
Update ocean from v3.5.9 to v3.5.11

jens-mueller-sociomantic:
* submodules/ocean v3.5.9(87446c2)...v3.5.11(c14340c) (6 commits)
  > Remove bogus out contract from IncrementalValue dtor.
  > Fix bug in Lzo.decompressSafe
  > Switch the FileSystemEvent test to use TaskEvent
  > Monitor for file changes inside MAKD_TMP, not inside /tmp
  > Unwatch the directory after the fileCreationTest has completed
  (...)


joseph-wakeling-sociomantic:
For reference, folks, this is at least potentially a bit of a problematic change, since it's not just a patch release, but a new minor release.  That creates potential breaking changes for downstreams of `dmxnet`.

Given that we're still in the v0.x.x release series, this is tolerable, but we should try to be conscious of what we're doing when we upgrade to a new minor (or major!) ocean release.  This isn't an application where we can just upgrade casually.

joseph-wakeling-sociomantic:
Since this hasn't been merged yet, can the dependency be switched to v3.5.11 (latest 3.5.x release)?

jens-mueller-sociomantic:
Can do.

joseph-wakeling-sociomantic (approved):

PR #51, opened 2018-01-31
Update base docker image to dlang:v4

joseph-wakeling-sociomantic:
This new dlang base image provides tangort 1.8.x and DMD 2.078.x.  This should allow us to make progress with dmxnet support for multithreading MXNet engines.

Part of sociomantic-tsunami/dmxnet#53.

### DIFF
--- a/beaver.Dockerfile
+++ b/beaver.Dockerfile
@@ -1,3 +1,3 @@
-FROM sociomantictsunami/dlang:v3
+FROM sociomantictsunami/dlang:v4
 COPY docker/ /docker-tmp
 RUN /docker-tmp/build && rm -fr /docker-tmp


joseph-wakeling-sociomantic:
Given the DMD update, this might require beaver v0.3.3 in order to work.  I'll submit that in a separate PR.

joseph-wakeling-sociomantic:
Weirdly, this seems to stall on `production` builds rather than D2 builds.  @leandro-lucarella-sociomantic are we missing anything here (other than the beaver update)?

joseph-wakeling-sociomantic:
Rebased on latest v0.x.x, which has updated to beaver v0.3.3.

This should not actually make a difference (it looks like the failures were in timeouts to pull the base docker images, not anything wrong on our side) but can't hurt.

joseph-wakeling-sociomantic:
Either way, now it works.  @jens-mueller-sociomantic this is probably most of interest to you given the wish to remove the `NaiveEngine` requirement, so I'll ask you to review it ;-)

jens-mueller-sociomantic:
I suggest adding `Part of sociomantic-tsunami/dmxnet#53`.

jens-mueller-sociomantic (changes requested):
I only suggest updating the commit message and PR description. Otherwise LGTM.

jens-mueller-sociomantic:
Please don't merge this yet, I want to figure out was going on in sociomantic-tsunami/dmxnet#54.

joseph-wakeling-sociomantic:
Updated commit message and PR description.  I don't think there's any problem in taking this in, BTW.  I don't think the problem with https://github.com/sociomantic-tsunami/dmxnet/pull/54 is on the dmxnet side (we need to first of all make sure the `libmxnet` package isn't responsible).

jens-mueller-sociomantic:
Hmm. Is there a way to verify that for making multi-threaded MXNet work, we need actually two changes? This one (changing the runtime) and the synchonizations introduced in sociomantic-tsunami/dmxnet#54. We can verify this locally of course.

joseph-wakeling-sociomantic:
You want to verify that this is actually a necessary change, you mean?

jens-mueller-sociomantic:
Yes. I want to verify that this is part of fixing multi-threading support.

joseph-wakeling-sociomantic:
Well, you already discovered that tangort 1.8.x was required to support this, no ... ?  And the `dlang:v3` images only use tangort 1.7.x:
https://github.com/sociomantic-tsunami/cachalot/blob/v3.2.0/dlang.Dockerfile

jens-mueller-sociomantic:
Yes, it's just extra verification.

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic do I understand right that you verified this now?  If so can we take this in?

jens-mueller-sociomantic:
No, not yet. The D2 build wasn't working for me locally.

jens-mueller-sociomantic:
The `ThreadedEngine` is not deadlocking with the old run-time. I think I need a `libmxnet` build that does perform actual threading. In D2 the integrationtest for the `ThreadedEnginePerDevice` dead locks. Don't know why this is. This is for the old run-time which I wasn't expecting.

For the new runtime introduced by this PR the integrationtest for the `ThreadedEnginePerDevice` still dead locks in D2. In D1 there is no dead lock. Interestingly, this is not the case for the Travis builds.
Can you reproduce the above results?

jens-mueller-sociomantic:
Just realized why it always works in Travis: Because we always use the `NaiveEngine`.

joseph-wakeling-sociomantic:
> I think I need a libmxnet build that does perform actual threading.

Yes.  If you don't have this then the other engines won't work.  I tested locally with a manually-built package that didn't disable it, and all worked fine.

jens-mueller-sociomantic:
> I tested locally with a manually-built package that didn't disable it, and all worked fine.

Did you remove the restriction to the `NaiveEngine` in the `Build.mak`?
And you could not reproduce any dead lock?

joseph-wakeling-sociomantic:
> Did you remove the restriction to the NaiveEngine in the `Build.mak`?

Yes, I swapped it out manually when running the tests.

>  And you could not reproduce any dead lock?

IIRC the `ThreadedEnginePerDevice` locked at the end of the program run.

FWIW this is feeling like a bit of a distraction in the short term.  There are a bunch of pieces that we need in place in order to be able to make this work correctly (starting with an updated MXNet package that supports multithreading).

Let's come back to things from the dmxnet side when we have everything we need from the MXNet side of things in place.

joseph-wakeling-sociomantic:
However ... in the meanwhile, can we get this PR in? :-)

jens-mueller-sociomantic:
I would like to confirm that this PR together with changes in sociomantic-tsunami/dmxnet#54 support properly fix the threading issues with a `libmxnet` supporting threaded engines. We can merge this. But it would make it harder to see the behavior on the old run-time on Travis. For this reason I would keep this open. When we have all parts we can take them as needed. Is there any reason to merge this now?

Why do you mention in the commit message the official D2 compiler version (`DMD 2.078.x`). Don't we only care about `dmd-transitional`?

joseph-wakeling-sociomantic:
Rebased on master.  @jens-mueller-sociomantic do you still want to hold off on this for now?

jens-mueller-sociomantic:
Since the deadlock issue seems to be fixed by MXNet 1.0.0 we can consider this independent. Either fix seems good.

jens-mueller-sociomantic:
Can you update the commit message and PR title?
I think it is misleading to mention the official D2 compiler version (`DMD 2.078.x`) since we only care about `dmd-transitional`.

joseph-wakeling-sociomantic:
Rebased on master and updated the commit message.

jens-mueller-sociomantic:
What about the `Part of sociomantic-tsunami/dmxnet#53.`? Does it still make sense?

joseph-wakeling-sociomantic:
Fair enough, I've removed it from the commit.  I'm leaving it in the PR description, though, since the discussion is relevant.

jens-mueller-sociomantic (approved):

PR #54, opened 2018-02-01
Provide support and CI tests for all MXNet engine types

jens-mueller-sociomantic:
This PR addresses synchronization issues of `NDArray.data` by two changes. The first commit replaces all write to slices obtained by `NDArray.data` by calls using the synchronized `NDArray.copyFrom`. The second commit make `data` return a read-only slice that is synchronized when the method executes. 
This addresses all synchronization issues when using `NDArray.data` in a multi-threaded setup. Hence, the restriction to the `NaiveEngine` is removed from `Build.mak` in the third commit (since this should be the only synchronization issue within `dmxnet` when using a multi-theaded MXNet engine). The fourth commit provides a target to test for all engines. The fifth and last commit instructs Travis to test with all MXNet engines to catch problem with engines during continuous integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,23 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
+        - BEAVER_DOCKER_VARS="MXNET_ENGINE_TYPE"
     matrix:
-        - DMD=dmd1 DIST=xenial F=production
-        - DMD=dmd1 DIST=xenial F=devel
-        - DMD=dmd-transitional DIST=xenial F=production
-        - DMD=dmd-transitional DIST=xenial F=devel
+        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
+        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
+        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
+
+        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
+        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
+        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
+
+        - DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
+        - DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
+        - DMD=dmd-transitional DIST=xenial F=production ThreadedEnginePerDevice
+
+        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
+        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
+        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
 
 install:
     - beaver dlang install
--- a/Build.mak
+++ b/Build.mak
@@ -8,10 +8,6 @@ ifeq ($(DVER),1)
 	override DFLAGS += -v2 -v2=-static-arr-params -v2=-volatile
 endif
 
-# use NaiveEngine (non-threaded) MXNet engine since the default (threaded)
-# version dead locks
-%test: export MXNET_ENGINE_TYPE=NaiveEngine
-
 .PHONY: download-mnist
 download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
--- a/integrationtest/mxnet/main.d
+++ b/integrationtest/mxnet/main.d
@@ -127,8 +127,8 @@ void main (istring[] args)
         assert(labels_batch.length == batch_size);
 
         // set batch and ...
-        matrix_x.data()[] = images_batch;
-        vector_y.data()[] = labels_batch;
+        matrix_x.copyFrom(images_batch);
+        vector_y.copyFrom(labels_batch);
 
         // make a forward and a backward pass
         executor.forward();
@@ -173,7 +173,7 @@ size_t[] predict (Symbol model, NDArray!(float) matrix_w,
 
     scope matrix_x = new NDArray!(float)(cpuContext(), [num_images, num_pixels]);
     scope (exit) matrix_x.freeHandle();
-    matrix_x.data()[] = images;
+    matrix_x.copyFrom(images);
     scope vector_y = new NDArray!(float)(cpuContext(), [num_images]);
     scope (exit) vector_y.freeHandle();
 
--- a/src/mxnet/Executor.d
+++ b/src/mxnet/Executor.d
@@ -247,14 +247,14 @@ public class Executor (T)
         // initialize two dimensional input vector
         scope input = new NDArray!(float)(context, [1, 2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         // and label to 1
         scope label = new NDArray!(float)(context, [1], 1f);
         scope (exit) label.freeHandle();
         // and weights
         scope weights = new NDArray!(float)(context, [1, 2]);
         scope (exit) weights.freeHandle();
-        weights.data[] = [0, 0];
+        weights.copyFrom([0, 0]);
         auto inputs = [input, weights, label];
 
         // initialize gradient w.r.t. weights to zero
@@ -279,14 +279,14 @@ public class Executor (T)
                                     input.data[1] * weights.data[1]]);
 
         // test with different inputs
-        input.data[] = [3, 4];
-        label.data[] = [3];
+        input.copyFrom([3, 4]);
+        label.copyFrom([3]);
         executor.forward(ForwardPass.outputs);
         test!("==")(output.data(), [input.data[0] * weights.data[0] +
                                     input.data[1] * weights.data[1]]);
 
         // test with different weights
-        weights.data[] = [3, 4];
+        weights.copyFrom([3, 4]);
         executor.forward(ForwardPass.outputs);
         test!("==")(output.data(), [input.data[0] * weights.data[0] +
                                     input.data[1] * weights.data[1]]);
@@ -339,14 +339,14 @@ public class Executor (T)
         // initialize two dimensional input vector
         scope input = new NDArray!(float)(context, [1, 2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         // and label to 1
         scope label = new NDArray!(float)(context, [1], 1f);
         scope (exit) label.freeHandle();
         // and weights
         scope weights = new NDArray!(float)(context, [1, 2]);
         scope (exit) weights.freeHandle();
-        weights.data[] = [0, 0];
+        weights.copyFrom([0, 0]);
         auto inputs = [input, weights, label];
 
         // initialize gradient w.r.t. weights to zero
@@ -372,8 +372,8 @@ public class Executor (T)
                                               (output.data[0] - label.data[0]) * input.data[1]]);
 
         // test with different inputs
-        input.data[] = [3, 4];
-        label.data[] = [3];
+        input.copyFrom([3, 4]);
+        label.copyFrom([3]);
         executor.forward();
         executor.backward();
         // gradient w.r.t. to w
@@ -381,7 +381,7 @@ public class Executor (T)
                                               (output.data[0] - label.data[0]) * input.data[1]]);
 
         // test with different weights
-        weights.data[] = [3, 4];
+        weights.copyFrom([3, 4]);
         executor.forward();
         executor.backward();
         // gradient w.r.t. to w
@@ -465,7 +465,7 @@ public class Executor (T)
         auto context = cpuContext();
         scope input = new NDArray!(float)(context, [2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         auto gradients = [NDArray!(float).init];
         auto gradients_ops = [OpReq.null_op];
         scope executor = new Executor(context, model, [input], gradients, gradients_ops, []);
--- a/src/mxnet/NDArray.d
+++ b/src/mxnet/NDArray.d
@@ -708,8 +708,7 @@ public class NDArray (T)
             scope (exit) a.freeHandle();
             scope b = new NDArray(cpuContext(), [2, 1]);
             scope (exit) b.freeHandle();
-            T[] b_data = [1, 2];
-            b.data[] = b_data;
+            b.copyFrom([1, 2]);
             a -= b;
             T[] a_data = [1, 1, 1,
                           0, 0, 0];
@@ -721,8 +720,7 @@ public class NDArray (T)
             scope (exit) a.freeHandle();
             scope b = new NDArray(cpuContext(), [1, 3]);
             scope (exit) b.freeHandle();
-            T[] b_data = [1, 2, 3];
-            b.data[] = b_data;
+            b.copyFrom([1, 2, 3]);
             a -= b;
             T[] a_data = [2, 1, 0,
                           2, 1, 0];
--- a/src/mxnet/NDArray.d
+++ b/src/mxnet/NDArray.d
@@ -344,27 +344,29 @@ public class NDArray (T)
         `[1, 2, 3,
           4, 5, 6]`.
 
-        Note that it is only safe to read/write this data once any pending
-        read/writes are completed. In a multi-threaded setup (i.e. when using
-        a threaded MXNet engine) calling `waitToRead` ensures all writes to
-        this slice have been finished. Similarly, when writing elements of the
-        slice calling `waitToWrite` ensures that all previous read/writes of
-        this slice have finished. `copyFrom` and `copyTo` perform these
-        synchronizations implicitly under the hood when copying from or to a
-        buffer.
+        This function calls `waitToRead` before returning the data slice. This
+        means all pending writes (when calling this method) have finished. But
+        note that it is not synchronized for later writes. Call `waitToRead` as
+        needed to synchronize. Alternatively, instead of using this method you
+        can perform a copy via `copyTo`.
+
+        Note that you cannot change elements of the data slice: use `copyFrom`
+        if you wish to write new values for the n-dimensional array data.
 
         Returns:
-            the data slice of this n-dimensional array
+            a read-only data slice of this n-dimensional array
 
     ***************************************************************************/
 
-    public T[] data ()
+    public Const!(T)[] data ()
     in
     {
         assert(this.mxnet_ndarray.exists());
     }
     body
     {
+        // all pending writes must have been finished
+        this.waitToRead();
         T* ptr;
         this.mxnet_ndarray.apply!(MXNDArrayGetData)(cast(void**) &ptr);
         return ptr[0 .. this.length];


jens-mueller-sociomantic:
Blocked by sociomantic-tsunami/dmxnet#51.

jens-mueller-sociomantic:
I wonder whether I should include a commit here updating the `.travis.yml` to test all MXNet engines.

jens-mueller-sociomantic:
Hmm. I was expecting that the D1 builds not using the `NaiveEngine` to fail (due to the missing tangort >=v1.8.0). At least the `ThreadedEnginePerDevice` fails but also for D2. To summarize, no D2 build should fail but all D1 builds not using the `NaiveEngine`. Once, sociomantic-tsunami/dmxnet#51 is merged. These should pass as well.

joseph-wakeling-sociomantic:
I had a similar experience last night trying things out, but of course there's a problem: our `libmxnet` package builds with multi-threading support disabled.  So no wonder things are not working ;-)

I'll try a fresh local build of the package (from scratch) and see whether that allows things to work.

jens-mueller-sociomantic:
It's annoying that you can specify an engine that won't work as advertised. I'll wait for your findings then.

joseph-wakeling-sociomantic:
Using a `libmxnet` with OpenMP support included allows these changes to work with all the different engines.  However, I'm a bit uncomfortable by the requirement to place all these manual wait calls throughout the code -- it seems like an easy way to introduce horrible bugs into a codebase.

Can we find some way to ensure that the wrapper takes care of the needed/expected `wait` calls without the user having to care?

BTW, on the data scale of our integration tests -- `NaiveEngine` is faster ;-)

.travis.yml
```diff
@@ -23,12 +23,22 @@ env:
...
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
-        - BEAVER_DOCKER_VARS="F"
+        - BEAVER_DOCKER_VARS="F MXNET_ENGINE_TYPE"
     matrix:
-        - DMD=dmd1 DIST=xenial F=production
-        - DMD=dmd1 DIST=xenial F=devel
-        - DMD=dmd-transitional DIST=xenial F=production
-        - DMD=dmd-transitional DIST=xenial F=devel
+        # Xenial D1 devel/production builds with all MXNet engines
+        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
```
jens-mueller-sociomantic: @leandro-lucarella-sociomantic Do you know of a way to avoid this explosion of combinations in the `.travis.yml`. Handling it inside `Build.mak` might also be an option.

jens-mueller-sociomantic:
Hmm. The `copyFrom` and `copyTo` perform the synchronization implicitly. But I'm not sure it's a good idea to always require using them and ban `data`. We could try to reduce the usage of `data` though.

integrationtest/mxnet/main.d
```diff
@@ -198,6 +198,7 @@ size_t[] predict (Symbol model, NDArray!(float) matrix_w,
     scope (exit) output.freeHandle();
     auto outputs = [output];
     executor.outputs(outputs);
+    output.waitToRead();
```
joseph-wakeling-sociomantic: This seems to me to be a case where we shouldn't have to manually wait: the `outputs` method should do it for us.

BTW I know it's only in integration tests, but we really should kill this practice of shadowy variable and method names.  `outputs(outputs)` is slightly disconcerting to read.

src/mxnet/Executor.d
```diff
@@ -260,6 +260,7 @@ public class Executor (T)
         // initialize gradient w.r.t. weights to zero
         scope gradient_weights = new NDArray!(float)(context, weights.shape(), 0f);
         scope (exit) gradient_weights.freeHandle();
+        gradient_weights.waitToRead();
```
joseph-wakeling-sociomantic: The wait here looks like something that could be handled in the constructor, no ... ?

src/mxnet/Executor.d
```diff
@@ -275,19 +276,22 @@ public class Executor (T)
 
         // update outputs only (gradients are updated only in backward pass)
         executor.forward(ForwardPass.outputs);
+        output.waitToRead();
```
joseph-wakeling-sociomantic: Ditto here this looks like something the `forward` method could handle under the hood (and similarly below).

src/mxnet/Executor.d
```diff
@@ -377,13 +384,17 @@ public class Executor (T)
         executor.forward();
         executor.backward();
         // gradient w.r.t. to w
+        output.waitToRead();
+        gradient_weights.waitToRead();
```
joseph-wakeling-sociomantic: This double wait seems particularly confusing and IIUC would be nicer handled under the hood. Is there any particular reason why one might want to issue forward & backward calls immediately and wait only after they were both issued?

integrationtest/mxnet/main.d
jens-mueller-sociomantic: Can't you have cases where you don't want to synchronize all the outputs? Maybe the caller is only accessing one.

src/mxnet/NDArray.d
```diff
@@ -166,6 +166,7 @@ public class NDArray (T)
         test!("==")(array.shape(), [2u, 2u]);
         test!("==")(array.length(), 2u * 2u);
         T[] data = [1, 1, 1, 1];
+        array.waitToRead();
```
joseph-wakeling-sociomantic: This seems unnecessary if the constructor handles the wait?

src/mxnet/Executor.d
jens-mueller-sociomantic: True. We could require that whenever we construct with values those values must have been written.

src/mxnet/NDArray.d
```diff
@@ -1030,6 +1042,7 @@ unittest
 
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret, lhs, rhs));
+                ret.waitToRead();
```
joseph-wakeling-sociomantic: All of these `applyWhateverOp` cases feel like ones where the wait could be done under the hood.

joseph-wakeling-sociomantic:
Looking through, I haven't been able to identify a case where it didn't feel like the `wait` requirement could be handled wrapper-side rather than putting it on the user.  Do you have any concrete examples otherwise ... ?

src/mxnet/Executor.d
jens-mueller-sociomantic: Does the `forward` know about the `outputs`?

integrationtest/mxnet/main.d
joseph-wakeling-sociomantic: My feeling is that the _default_ should always be to wait, and that maybe, if we have a use-case (only if!), we should offer a way to do a wait-free call.

In other words, the behaviour should be safe by default, wait-free on request (if we ever have a use-case where that makes sense).

src/mxnet/Executor.d
jens-mueller-sociomantic: I have to look into this case.

src/mxnet/Executor.d
joseph-wakeling-sociomantic: I don't know if there's a risk to the constructor if it calls `wait` methods directly (since IIRC they might throw).  But I don't consider that a _big_ risk.

src/mxnet/NDArray.d
jens-mueller-sociomantic: Should be. Having the constructor wait is an option. The only thing I'm not sure of is its impact on performance. I think when you synchronize but it not needed you introduce a dependency that may slow down execution.

src/mxnet/Executor.d
joseph-wakeling-sociomantic: Well, the `Executor` has to know about them somehow, surely?

jens-mueller-sociomantic:
For me it's the other way around. When we always synchronize it won't run in parallel when it could. It is only `data` that causes the problems.

jens-mueller-sociomantic:
Maybe we should just check our internal use of `data` and provide a `dataSync` method. Then `data` should be only used when you want to avoid the synchronization and it is safe. In the test for example we would use `dataSync` then.

integrationtest/mxnet/main.d
jens-mueller-sociomantic: The question is whether we have to keep the wait-free use case in mind. Because I don't know how to support wait-free when we perform the synchronizations in various parts of the code.

src/mxnet/Executor.d
jens-mueller-sociomantic: They are passed to the `outputs` method.

joseph-wakeling-sociomantic:
It sounds like what we really need is to have clarity about which methods involve the user either writing values to (or reading values from) the `NDArray` class, and all those methods should insert appropriate wait commands.

Regarding forward and backward passes: do I understand right that these are essentially scheduling actions, and so if I go `forward; backward` I'm guaranteed to get correctly ordered behaviour (rather than accidentally starting the backward pass "in parallel" to the forward one)?

jens-mueller-sociomantic:
> It sounds like what we really need is to have clarity about which methods involve the user either writing values to (or reading values from) the `NDArray` class, and all those methods should insert appropriate wait commands.

Instead of going through all these methods. Why not address the problem at its cause. The calculations themselves are already good. It is only when using `data` which is not synchronized. This is the cause of the problem. It is much simpler to add an `dataSync` and use that still having the unsynced version. Probably we should make `data` synced and call the other one `dataNotSync` or provide an overload to `data` with sync by default. I rather want to fix the problematic method than fix all of the existing. Because this introduces synchronizations when you may not need them and you have to fix more methods. What do you think?

> Regarding forward and backward passes: do I understand right that these are essentially scheduling actions, and so if I go `forward; backward` I'm guaranteed to get correctly ordered behaviour (rather than accidentally starting the backward pass "in parallel" to the forward one)?

Yes, that is my understanding.

joseph-wakeling-sociomantic:
Just to follow up based on discussion, it looks like we have the following ways forward:

* `data()` can add a sync call immediately prior to returning the slice (just to make sure that the data is valid)

* `data()` can switch to returning only a `Const!(T[])` so that the resulting slice and its elements cannot be modified (at least in D2).

This should at least deal with the most basic needs of our test cases.

jens-mueller-sociomantic:
Rebased and added new approach to synchronize `data` directly.

```diff
diff --git a/integrationtest/mxnet/main.d b/integrationtest/mxnet/main.d
index 54e93cf..ba17099 100644
--- a/integrationtest/mxnet/main.d
+++ b/integrationtest/mxnet/main.d
@@ -198,7 +198,6 @@ size_t[] predict (Symbol model, NDArray!(float) matrix_w,
     scope (exit) output.freeHandle();
     auto outputs = [output];
     executor.outputs(outputs);
-    output.waitToRead();
     auto predictions_vector = to!(float[])(output.data());
 
     auto predictions = new size_t[num_images];
diff --git a/src/mxnet/Executor.d b/src/mxnet/Executor.d
index 9dd0b7b..0d7cdcf 100644
--- a/src/mxnet/Executor.d
+++ b/src/mxnet/Executor.d
@@ -247,20 +247,19 @@ public class Executor (T)
         // initialize two dimensional input vector
         scope input = new NDArray!(float)(context, [1, 2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         // and label to 1
         scope label = new NDArray!(float)(context, [1], 1f);
         scope (exit) label.freeHandle();
         // and weights
         scope weights = new NDArray!(float)(context, [1, 2]);
         scope (exit) weights.freeHandle();
-        weights.data[] = [0, 0];
+        weights.copyFrom([0, 0]);
         auto inputs = [input, weights, label];
 
         // initialize gradient w.r.t. weights to zero
         scope gradient_weights = new NDArray!(float)(context, weights.shape(), 0f);
         scope (exit) gradient_weights.freeHandle();
-        gradient_weights.waitToRead();
         test!("==")(gradient_weights.data(), [0f, 0f]);
         auto gradients = [NDArray!(float).init, gradient_weights, NDArray!(float).init];
         auto gradients_ops = [OpReq.null_op, OpReq.write, OpReq.null_op];
@@ -276,22 +275,19 @@ public class Executor (T)
 
         // update outputs only (gradients are updated only in backward pass)
         executor.forward(ForwardPass.outputs);
-        output.waitToRead();
         test!("==")(output.data(), [input.data[0] * weights.data[0] +
                                     input.data[1] * weights.data[1]]);
 
         // test with different inputs
-        input.data[] = [3, 4];
-        label.data[] = [3];
+        input.copyFrom([3, 4]);
+        label.copyFrom([3]);
         executor.forward(ForwardPass.outputs);
-        output.waitToRead();
         test!("==")(output.data(), [input.data[0] * weights.data[0] +
                                     input.data[1] * weights.data[1]]);
 
         // test with different weights
-        weights.data[] = [3, 4];
+        weights.copyFrom([3, 4]);
         executor.forward(ForwardPass.outputs);
-        output.waitToRead();
         test!("==")(output.data(), [input.data[0] * weights.data[0] +
                                     input.data[1] * weights.data[1]]);
     }
@@ -343,20 +339,19 @@ public class Executor (T)
         // initialize two dimensional input vector
         scope input = new NDArray!(float)(context, [1, 2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         // and label to 1
         scope label = new NDArray!(float)(context, [1], 1f);
         scope (exit) label.freeHandle();
         // and weights
         scope weights = new NDArray!(float)(context, [1, 2]);
         scope (exit) weights.freeHandle();
-        weights.data[] = [0, 0];
+        weights.copyFrom([0, 0]);
         auto inputs = [input, weights, label];
 
         // initialize gradient w.r.t. weights to zero
         scope gradient_weights = new NDArray!(float)(context, weights.shape(), 0f);
         scope (exit) gradient_weights.freeHandle();
-        gradient_weights.waitToRead();
         test!("==")(gradient_weights.data(), [0f, 0f]);
         auto gradients = [NDArray!(float).init, gradient_weights, NDArray!(float).init];
         auto gradients_ops = [OpReq.null_op, OpReq.write, OpReq.null_op];
@@ -373,28 +368,22 @@ public class Executor (T)
         executor.forward();
         executor.backward();
         // gradient w.r.t. to w is (A*w'-b) * A
-        output.waitToRead();
-        gradient_weights.waitToRead();
         test!("==")(gradient_weights.data(), [(output.data[0] - label.data[0]) * input.data[0],
                                               (output.data[0] - label.data[0]) * input.data[1]]);
 
         // test with different inputs
-        input.data[] = [3, 4];
-        label.data[] = [3];
+        input.copyFrom([3, 4]);
+        label.copyFrom([3]);
         executor.forward();
         executor.backward();
         // gradient w.r.t. to w
-        output.waitToRead();
-        gradient_weights.waitToRead();
         test!("==")(gradient_weights.data(), [(output.data[0] - label.data[0]) * input.data[0],
                                               (output.data[0] - label.data[0]) * input.data[1]]);
 
         // test with different weights
-        weights.data[] = [3, 4];
+        weights.copyFrom([3, 4]);
         executor.forward();
         executor.backward();
-        output.waitToRead();
-        gradient_weights.waitToRead();
         // gradient w.r.t. to w
         test!("==")(gradient_weights.data(), [(output.data[0] - label.data[0]) * input.data[0],
                                               (output.data[0] - label.data[0]) * input.data[1]]);
@@ -476,7 +465,7 @@ public class Executor (T)
         auto context = cpuContext();
         scope input = new NDArray!(float)(context, [2]);
         scope (exit) input.freeHandle();
-        input.data[] = [1, 2];
+        input.copyFrom([1, 2]);
         auto gradients = [NDArray!(float).init];
         auto gradients_ops = [OpReq.null_op];
         scope executor = new Executor(context, model, [input], gradients, gradients_ops, []);
@@ -485,7 +474,6 @@ public class Executor (T)
         scope (exit) output.freeHandle();
         scope outputs = [output];
         executor.outputs(outputs);
-        output.waitToRead();
         test!("==")(output.data, input.data);
 
         scope output2 = new NDArray!(float)(null);
diff --git a/src/mxnet/NDArray.d b/src/mxnet/NDArray.d
index ad64d9c..2d7d7c3 100644
--- a/src/mxnet/NDArray.d
+++ b/src/mxnet/NDArray.d
@@ -166,7 +166,6 @@ public class NDArray (T)
         test!("==")(array.shape(), [2u, 2u]);
         test!("==")(array.length(), 2u * 2u);
         T[] data = [1, 1, 1, 1];
-        array.waitToRead();
         test!("==")(array.data(), data);
     }
 
@@ -345,27 +344,30 @@ public class NDArray (T)
         `[1, 2, 3,
           4, 5, 6]`.
 
-        Note that it is only safe to read/write this data once any pending
-        read/writes are completed. In a multi-threaded setup (i.e. when using
-        a threaded MXNet engine) calling `waitToRead` ensures all writes to
-        this slice have been finished. Similarly, when writing elements of the
-        slice calling `waitToWrite` ensures that all previous read/writes of
-        this slice have finished. `copyFrom` and `copyTo` perform these
-        synchronizations implicitly under the hood when copying from or to a
-        buffer.
+        This function calls `waitToRead` implicitly before returning the data
+        slice. This means all pending writes (when calling this method) have
+        finished. But note that it is not synchronized for later writes.
+        Meaning when you keep the slice around for later access you need to
+        call `waitToRead` as needed.
+
+        Note that you cannot change elements of the data slice. You can use
+        `copyFrom` changing the all elements. Changing individual elements is
+        not supported but might be added in future.
 
         Returns:
-            the data slice of this n-dimensional array
+            a read-only data slice of this n-dimensional array
 
     ***************************************************************************/
 
-    public T[] data ()
+    public Const!(T)[] data ()
     in
     {
         assert(this.mxnet_ndarray.exists());
     }
     body
     {
+        // all pending writes must have been finished
+        this.waitToRead();
         T* ptr;
         this.mxnet_ndarray.apply!(MXNDArrayGetData)(cast(void**) &ptr);
         return ptr[0 .. this.length];
@@ -377,7 +379,6 @@ public class NDArray (T)
         scope (exit) array.freeHandle();
         T[] data = [1, 1, 1,
                     1, 1, 1];
-        array.waitToRead();
         test!("==")(array.data(), data);
     }
 
@@ -483,7 +484,6 @@ public class NDArray (T)
         T[] data = [1, 2, 3,
                     4, 5, 6];
         a.copyFrom(data);
-        a.waitToRead();
         test!("==")(a.data(), data);
     }
 
@@ -643,12 +643,10 @@ public class NDArray (T)
         testNoAlloc(a = 1);
         T[] data = [1, 1, 1,
                     1, 1, 1];
-        a.waitToRead();
         test!("==")(a.data(), data);
         data = [4, 4, 4,
                 4, 4, 4];
         testNoAlloc(a = 4);
-        a.waitToRead();
         test!("==")(a.data(), data);
     }
 
@@ -694,7 +692,6 @@ public class NDArray (T)
             a -= b;
             T[] a_data = [8, 8, 8,
                           8, 8, 8];
-            a.waitToRead();
             test!("==")(a.data(), a_data);
         }
         // subtract scalar b
@@ -706,7 +703,6 @@ public class NDArray (T)
             a -= b;
             T[] a_data = [7, 7, 7,
                           7, 7, 7];
-            a.waitToRead();
             test!("==")(a.data(), a_data);
         }
         // column-wise subtract column vector b
@@ -715,12 +711,10 @@ public class NDArray (T)
             scope (exit) a.freeHandle();
             scope b = new NDArray(cpuContext(), [2, 1]);
             scope (exit) b.freeHandle();
-            T[] b_data = [1, 2];
-            b.data[] = b_data;
+            b.copyFrom([1, 2]);
             a -= b;
             T[] a_data = [1, 1, 1,
                           0, 0, 0];
-            a.waitToRead();
             test!("==")(a.data(), a_data);
         }
         // row-wise subtract row vector b
@@ -729,12 +723,10 @@ public class NDArray (T)
             scope (exit) a.freeHandle();
             scope b = new NDArray(cpuContext(), [1, 3]);
             scope (exit) b.freeHandle();
-            T[] b_data = [1, 2, 3];
-            b.data[] = b_data;
+            b.copyFrom([1, 2, 3]);
             a -= b;
             T[] a_data = [2, 1, 0,
                           2, 1, 0];
-            a.waitToRead();
             test!("==")(a.data(), a_data);
         }
     }
@@ -772,17 +764,14 @@ public class NDArray (T)
         scope (exit) a.freeHandle();
         T[] a_data = [1, 1, 1,
                       1, 1, 1];
-        a.waitToRead();
         test!("==")(a.data(), a_data);
         a_data = [2, 2, 2,
                   2, 2, 2];
         a *= to!(T)(2);
-        a.waitToRead();
         test!("==")(a.data(), a_data);
         a_data = [6, 6, 6,
                   6, 6, 6];
         a *= to!(T)(3);
-        a.waitToRead();
         test!("==")(a.data(), a_data);
     }
 
@@ -1042,7 +1031,6 @@ unittest
 
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret, lhs, rhs));
-                ret.waitToRead();
                 test!("==")(ret.data(),
                             elementWiseTestOp!(op)(lhs_data, rhs_data));
             }
@@ -1071,7 +1059,6 @@ unittest
 
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret, lhs, rhs));
-                ret.waitToRead();
                 test!("==")(ret.data(),
                             elementWiseTestOp!(op)(lhs_data, rhs_data));
             }
@@ -1104,7 +1091,6 @@ unittest
                 auto expected_data = elementWiseTestOp!(op)(lhs_data,
                                                             rhs_broadcast_data);
 
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
 
                 // validate consistency of `applyBroadcastOp` results
@@ -1117,7 +1103,6 @@ unittest
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret_b, lhs, rhs_b));
 
-                ret_b.waitToRead();
                 test!("==")(ret_b.data(), expected_data);
             }
 
@@ -1150,7 +1135,6 @@ unittest
                 auto expected_data = elementWiseTestOp!(op)(lhs_broadcast_data,
                                                             rhs_data);
 
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
 
                 // validate consistency of `applyBroadcastOp` results
@@ -1163,7 +1147,6 @@ unittest
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret_b, lhs_b, rhs));
 
-                ret_b.waitToRead();
                 test!("==")(ret_b.data(), expected_data);
 
             }
@@ -1203,7 +1186,6 @@ unittest
                 auto expected_data = elementWiseTestOp!(op)(lhs_broadcast_data,
                                                             rhs_broadcast_data);
 
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
 
                 // validate consistency of `applyBroadcastOp` results
@@ -1221,7 +1203,6 @@ unittest
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret_b, lhs_b, rhs_b));
 
-                ret_b.waitToRead();
                 test!("==")(ret_b.data(), expected_data);
             }
 
@@ -1260,7 +1241,6 @@ unittest
                 auto expected_data = elementWiseTestOp!(op)(lhs_broadcast_data,
                                                             rhs_broadcast_data);
 
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
 
                 // validate consistency of `applyBroadcastOp` results
@@ -1278,7 +1258,6 @@ unittest
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret_b, lhs_b, rhs_b));
 
-                ret_b.waitToRead();
                 test!("==")(ret_b.data(), expected_data);
             }
 
@@ -1308,7 +1287,6 @@ unittest
 
                 T expected_val = scalarTestOp!(op)(lhs_val, rhs_val);
 
-                ret.waitToRead();
                 foreach (val; ret.data())
                 {
                     test!("==")(val, expected_val);
@@ -1340,7 +1318,6 @@ unittest
                 assert(ret.data() != expected_data);
 
                 applyBroadcastOp(broadcastOperator(op), ret, ret, ret);
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
                 test!("!=")(ret.data(), lhs_data);
 
@@ -1355,7 +1332,6 @@ unittest
                 assert(other.data() == rhs_data);
 
                 applyBroadcastOp(broadcastOperator(op), ret, ret, other);
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
                 test!("!=")(ret.data(), lhs_data);
                 test!("==")(other.data(), rhs_data);
@@ -1369,7 +1345,6 @@ unittest
                 assert(other.data() == lhs_data);
 
                 applyBroadcastOp(broadcastOperator(op), ret, other, ret);
-                ret.waitToRead();
                 test!("==")(ret.data(), expected_data);
                 test!("!=")(ret.data(), rhs_data);
                 test!("==")(other.data(), lhs_data);
@@ -1492,8 +1467,6 @@ unittest
                 // compare to results of already tested broadcast operation
                 testNoAlloc(applyBroadcastOp(broadcastOperator(op),
                                              ret_broadcast, lhs, rhs_broadcast));
-                ret.waitToRead();
-                ret_broadcast.waitToRead();
                 test!("==")(ret.data(), ret_broadcast.data());
                 test!("==")(lhs.data(), lhs_data);
 
@@ -1517,8 +1490,6 @@ unittest
                 assert(ret_in_place.data().ptr !is lhs_data.ptr);
 
                 testNoAlloc(applyScalarOp(op, ret_in_place, ret_in_place, rhs));
-                ret_in_place.waitToRead();
-                ret_broadcast.waitToRead();
                 test!("==")(ret_in_place.data(), ret_broadcast.data());
 
                 if (rhs == op_identity)
```

joseph-wakeling-sociomantic:
I suggest dropping the Travis changes for now.  The changes to the `data` method are fine as they are and we can follow up later if needed.

jens-mueller-sociomantic:
You mean the Travis changes **and** the change in `Build.mak`?

joseph-wakeling-sociomantic:
Yes, the entire `Make Travis test all MXNet engines` patch can be dropped for now.

FWIW if we were going to pursue this, I would suggest listing the options so that the `DIST` and `F` options are all grouped together for the same engine:

```yaml
matrix:
- DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
- DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
- DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice

- DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
- DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
- DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEnginePerDevice

- DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
- DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
- DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice

- DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
- DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
- DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
```

... but I'd still rather set things up in a way so that integration tests will always cover all the different engine types.

jens-mueller-sociomantic:
```diff
diff --git a/.travis.yml b/.travis.yml
index 9c9d2a7..27ada0e 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,22 +23,12 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
-        - BEAVER_DOCKER_VARS="F MXNET_ENGINE_TYPE"
+        - BEAVER_DOCKER_VARS="F"
     matrix:
-        # Xenial D1 devel/production builds with all MXNet engines
-        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
-        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
-        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
-        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
-        - DMD=dmd1 DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
-        - DMD=dmd1 DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
-        # Xenial D2 devel/production builds with all MXNet engines
-        - DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=NaiveEngine
-        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=NaiveEngine
-        - DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEngine
-        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEngine
-        - DMD=dmd-transitional DIST=xenial F=production MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
-        - DMD=dmd-transitional DIST=xenial F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
+        - DMD=dmd1 DIST=xenial F=production
+        - DMD=dmd1 DIST=xenial F=devel
+        - DMD=dmd-transitional DIST=xenial F=production
+        - DMD=dmd-transitional DIST=xenial F=devel
 
 install:
     - beaver dlang install
```

joseph-wakeling-sociomantic:
Minor: suggest being explicit about it being `NDArray.data` in the 1st commit message.  (If you need to save characters in the title, drop the backtick marks;-)

joseph-wakeling-sociomantic:
Also, in the 2nd commit message, it's only `Part of` the issue, it doesn't fix it by itself ;-)

joseph-wakeling-sociomantic:
You can however add that cross-ref to the PR description.

jens-mueller-sociomantic:
Why do you think the second commit does not fix the issue? I don't see what is left (assuming we merge the `Dockerfile` with version 4). We only need a `libmxnet` with multi-thread support to test.

joseph-wakeling-sociomantic:
> Why do you think the second commit does not fix the issue?

The issue isn't fixed until all the different pieces are addressed and we have confirmed that it really is fixed.  Since some of those aspects are out of MXNet's hands (e.g. the `libmxnet` we use), it's best to not declare the commit as a fix.

jens-mueller-sociomantic:
Alright. Then this is a misunderstanding. I want to merge this only when the other issues (`Dockerfile` update and `libmxnet` with threaded engines) have been addressed. That's why I also wanted to verify the issue using Travis.

joseph-wakeling-sociomantic:
OK, as you wish.  Let's hold up on these factors then.

jens-mueller-sociomantic:
With the prerelease of libmxnet 1.0.0 this is not blocked anymore as it fixes deadlocks with threaded engines. The fix is provided within mxnet itself. An alternative solution is provided by sociomantic-tsunami/dmxnet#51. Either one is good. That's why not blocking anymore on sociomantic-tsunami/dmxnet#51.

jens-mueller-sociomantic:
I'd like to remove the removal of the `NaiveEngine` and make sure that all engines are tested though.

jens-mueller-sociomantic:
```diff
diff --git a/.travis.yml b/.travis.yml
index 27ada0e..fe60a8b 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,4 +34,6 @@ install:
     - beaver dlang install
     - beaver run script/download-mnist
 
-script: beaver dlang make
+script:
+    - beaver dlang make
+    - beaver dlang make test-all-engines
diff --git a/Build.mak b/Build.mak
index 80693b8..e2a3c8b 100644
--- a/Build.mak
+++ b/Build.mak
@@ -21,3 +21,10 @@ $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
 $O/test-mxnet.stamp: download-mnist
 
 $O/%unittests: override LDFLAGS += -lz
+
+# always run tests with each engine passing current environment
+.PHONY: test-all-engines
+test-all-engines:
+	$(MAKE) $(MAKEFLAGS) -B MXNET_ENGINE_TYPE=NaiveEngine test
+	$(MAKE) $(MAKEFLAGS) -B MXNET_ENGINE_TYPE=ThreadedEngine test
+	$(MAKE) $(MAKEFLAGS) -B MXNET_ENGINE_TYPE=ThreadedEnginePerDevice test
```

jens-mueller-sociomantic:
Rebased.

beaver.Dockerfile
```diff
@@ -1,3 +1,7 @@
 FROM sociomantictsunami/dlang:v3
 COPY docker/ /docker-tmp
 RUN /docker-tmp/build && rm -fr /docker-tmp
+
+# temporary fix to use 1.0.0 libmxnet
+RUN sudo apt-get update && sudo apt-get -y install \
+        libmxnet='1.0.0-*'
```
joseph-wakeling-sociomantic: This won't work without an update to `docker/build` to update the mxnet repo settings (you need `prerelease`).

joseph-wakeling-sociomantic: Note, enabling `prerelease` is only a short-term workaround, we should transition the package to the `release` repo before finalizing this.

jens-mueller-sociomantic:
I am getting

```
[14:16:08] src/engine/engine.cc:55: MXNet start using engine: ThreadedEnginePerDevice
terminate called after throwing an instance of 'std::system_error'
  what():  Operation not permitted
```

for `make -B MXNET_ENGINE_TYPE=ThreadedEnginePerDevice unittest` from time to time. Strangely make still exists with success which makes it easy to overlook.

jens-mueller-sociomantic:
Seems to be apache/incubator-mxnet#9213.

joseph-wakeling-sociomantic:
Hmmm, let's see if an update to MXNet 1.1.0 will fix this.

jens-mueller-sociomantic:
Blocking this until we have a 1.1.0 MXNet.

jens-mueller-sociomantic:
@joseph-wakeling-sociomantic Can you build a 1.1.0 (pre)release while this issue is still fresh in my mind?

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic you should be able to build your own from https://github.com/sociomantic-tsunami/mxnet/pull/27:

```
make -jN USE_BLAS=openblas USE_OPENCV=0
F=production ./build-tsunami-package
```

I tried that now with a checkout of this PR and it seemed to work with all the engines, with no deadlocks and all tests passing.

Build.mak
```diff
@@ -25,3 +21,10 @@ $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
 $O/test-mxnet.stamp: download-mnist
 
 $O/%unittests: override LDFLAGS += -lz
+
+# always run tests with each engine passing current environment
+.PHONY: test-all-engines
+test-all-engines:
+	$(MAKE) -$(MAKEFLAGS) -B MXNET_ENGINE_TYPE=NaiveEngine test
+	$(MAKE) -$(MAKEFLAGS) -B MXNET_ENGINE_TYPE=ThreadedEngine test
+	$(MAKE) -$(MAKEFLAGS) -B MXNET_ENGINE_TYPE=ThreadedEnginePerDevice test
```
joseph-wakeling-sociomantic: Placing `$(MAKEFLAGS)` here won't work.  It needs to be after the `-B ...` part and before `test`.

But I'm not sure this is the right way to go about things.  It might be better to define a default engine value, which can be overridden by the user.

jens-mueller-sociomantic: What kind of failure are you seeing? Because it works for me.

jens-mueller-sociomantic: > But I'm not sure this is the right way to go about things. It might be better to define a default engine value, which can be overridden by the user.

This does already exist. You can use `MXNET_ENGINE_TYPE`.

jens-mueller-sociomantic:
I am going to wait for sociomantic-tsunami/mxnet release because will need it anyway for Travis to pass.

joseph-wakeling-sociomantic:
I'd really appreciate it if before that you could give a go at building and installing a package from https://github.com/sociomantic-tsunami/mxnet/pull/27 and checking if it makes a difference to you.

There are still a few niggles to be ironed out with the packaging and deb repo management, so before having to push anyone on that, I'd like to make sure that the new package is actually going to address our issue.

Build.mak
joseph-wakeling-sociomantic: > What kind of failure are you seeing? Because it works for me.

```
$ make d2conv
$ make DVER=2 test-all-engines 
!! Flags or commands changed: D compiler; re-building affected files...
make - -- DVER=2 -B MXNET_ENGINE_TYPE=NaiveEngine test
make[1]: *** No rule to make target '-B'. Stop.
Build.mak:28: recipe for target 'test-all-engines' failed
make: *** [test-all-engines] Error 2
```

The hardcoded `-` before `$(MAKEFLAGS)` is causing trouble, just for starters.

> This does already exist. You can use `MXNET_ENGINE_TYPE`.

Yes, and that's the point.  The CI can just define multiple jobs with the different engines.

Locally speaking, there's no particular trouble to do either separate `make test` calls or to add a `test-all-engines` target to a personal `Config.local.mak`.  But it seems overkill to add it to the main `Build.mak`.

jens-mueller-sociomantic: Defining multiple configurations is annoying in Travis. I can add them as I had. I went with this version because it's simpler and can be used to easily verify locally whether a change works for all engines. I'll update the `.travis.yml` then accordingly and remove the `test-all-engines` target.

joseph-wakeling-sociomantic: I don't suggest acting one way or another right now.  Let's get some others in to advise on the best way to achieve the wanted outcome.

jens-mueller-sociomantic: Okay. Then let's wait until this is settled.

jens-mueller-sociomantic:
I can confirm that the package I build based on sociomantic-tsunami/mxnet#27 works. The permissions of the installed files seemed a bit odd but likely to my locally configured `umask`. `/usr/lib/libmxnet.so` was `rwx` for user and nothing else. `/usr/lib/libmxnet.a` was `rw` for user and nothing else. So I changed to include `r` for group and others. Otherwise it won't find the library for linking. Again, this is probably due to my local `umask`. I don't know whether the package should be explicit about it.

jens-mueller-sociomantic:
Rebased.

jens-mueller-sociomantic:
I can confirm that it works locally when using the 1.1.0 release.

jens-mueller-sociomantic:
Pushed using libmxnet 1.1.0 to check if it works on Travis.

joseph-wakeling-sociomantic:
@jens-mueller-sociomantic you may need to ensure the `prerelease` repo is added (just temporarily).

jens-mueller-sociomantic:
I just saw it. I thought the mxnet package would be in the release repository.

jens-mueller-sociomantic:
Assuming this passes we are left with deciding how to integrate testing of the new engines.

jens-mueller-sociomantic:
Removed prerelease libmxnet, i.e., use latest release libmxnet which should be
1.1.0.

jens-mueller-sociomantic:
@joseph-wakeling-sociomantic I think the only thing left here is to decide how to integrate testing of the different engines.

joseph-wakeling-sociomantic:
Right.  I don't remember all the details of our discussion, but looking at the current patch to `.travis.yml` it strikes me that this is an unnecessary volume of CI work: for any given `DMD`, `DIST` and `F` combo, we'd ideally like to test _all_ the engines from one build.

OTOH having separate entries like this means that unittests will get run as well.  The waste is in the fact that unittests and integration tests are rebuilt multiple times rather than being built once and just _run_ multiple times with the different engines.

@leandro-lucarella-sociomantic I wonder if you can advise whether makd has any options to support us here.  The ideal that would happen is that we'd specify our builds with:

```
DMD=... DIST=... F=...
```

... and then the unittests and integration tests would be built once, but then run 3 different times with `MXNET_ENGINE_TYPE={NaiveEngine,ThreadedEngine,ThreadedEnginePerDevice}`.

joseph-wakeling-sociomantic:
(However, if the currently-proposed changes to `.travis.yml` let us move forward, I'm happy to accept them as a short-term compromise.)

joseph-wakeling-sociomantic:
(And yes, I recognize that the current travis-jobs-based approach was one I suggested, but that was in reaction to `Build.mak` setup that used internal make calls to run the different test setups.  The CI job approach is arguably better than that, but it'd be good to see if we can find an alternative approach.)

src/mxnet/NDArray.d
```diff
@@ -344,27 +344,30 @@ public class NDArray (T)
...
-        synchronizations implicitly under the hood when copying from or to a
-        buffer.
+        This function calls `waitToRead` implicitly before returning the data
+        slice. This means all pending writes (when calling this method) have
+        finished. But note that it is not synchronized for later writes.
+        Meaning when you keep the slice around for later access you need to
+        call `waitToRead` as needed.
+
+        Note that you cannot change elements of the data slice. You can use
+        `copyFrom` changing the all elements. Changing individual elements is
```
joseph-wakeling-sociomantic: This documentation reads a bit oddly.  How about

```
Note that you cannot change elements of the data slice: use `copyFrom`
if you wish to write new values for the n-dimensional array data.
```

src/mxnet/NDArray.d
```diff
@@ -344,27 +344,30 @@ public class NDArray (T)
...
-        this slice have been finished. Similarly, when writing elements of the
-        slice calling `waitToWrite` ensures that all previous read/writes of
-        this slice have finished. `copyFrom` and `copyTo` perform these
-        synchronizations implicitly under the hood when copying from or to a
-        buffer.
+        This function calls `waitToRead` implicitly before returning the data
+        slice. This means all pending writes (when calling this method) have
+        finished. But note that it is not synchronized for later writes.
+        Meaning when you keep the slice around for later access you need to
+        call `waitToRead` as needed.
```
joseph-wakeling-sociomantic: Isn't the concern here more that the values in the slice might change unexpectedly?  Might be worth referencing `copyTo` as a way to make a reliable deep-copy that will _not_ be updated unexpectedly in-place.

src/mxnet/NDArray.d
```diff
@@ -344,27 +344,30 @@ public class NDArray (T)
...
 
-        Note that it is only safe to read/write this data once any pending
-        read/writes are completed. In a multi-threaded setup (i.e. when using
-        a threaded MXNet engine) calling `waitToRead` ensures all writes to
-        this slice have been finished. Similarly, when writing elements of the
-        slice calling `waitToWrite` ensures that all previous read/writes of
-        this slice have finished. `copyFrom` and `copyTo` perform these
-        synchronizations implicitly under the hood when copying from or to a
-        buffer.
+        This function calls `waitToRead` implicitly before returning the data
```
joseph-wakeling-sociomantic: I don't quite follow what is meant by `implicitly` here.  It quite explicitly calls it!

src/mxnet/NDArray.d
jens-mueller-sociomantic: I wanted to emphasize here that you can keep the slice around but you have to synchronize it yourself. Of course you can also use `copyTo` to make a copy. It's another option at the expense of copying.

src/mxnet/NDArray.d
jens-mueller-sociomantic: Seems it is misleading. I remove it.

jens-mueller-sociomantic:
```diff
diff --git a/src/mxnet/NDArray.d b/src/mxnet/NDArray.d
index 2d7d7c3..97e600c 100644
--- a/src/mxnet/NDArray.d
+++ b/src/mxnet/NDArray.d
@@ -344,15 +344,14 @@ public class NDArray (T)
         `[1, 2, 3,
           4, 5, 6]`.
 
-        This function calls `waitToRead` implicitly before returning the data
-        slice. This means all pending writes (when calling this method) have
-        finished. But note that it is not synchronized for later writes.
-        Meaning when you keep the slice around for later access you need to
-        call `waitToRead` as needed.
+        This function calls `waitToRead` before returning the data slice. This
+        means all pending writes (when calling this method) have finished. But
+        note that it is not synchronized for later writes. Call `waitToRead` as
+        needed to synchronize. Alternatively, instead of using this method you
+        can perform a copy via `copyTo`.
 
-        Note that you cannot change elements of the data slice. You can use
-        `copyFrom` changing the all elements. Changing individual elements is
-        not supported but might be added in future.
+        Note that you cannot change elements of the data slice: use `copyFrom`
+        if you wish to write new values for the n-dimensional array data.
 
         Returns:
             a read-only data slice of this n-dimensional array
```

jens-mueller-sociomantic:
Rebased.

joseph-wakeling-sociomantic (commented):
If you can drop the `[WIP]` from the `.travis.yml` patch, I think this is good to go :-)

joseph-wakeling-sociomantic:
(Please also rebase to get the latest ocean and makd updates.)

jens-mueller-sociomantic:
> If you can drop the `[WIP]` from the `.travis.yml` patch, I think this is good to go :-)

How are we going to test the new engines?

joseph-wakeling-sociomantic:
> How are we going to test the new engines?

Doesn't the new Travis setup do that?  It seems adequate for the short term (albeit it would be nice to have a test-everything option for a single build).

jens-mueller-sociomantic:
Ah. I didn't get it. You're fine with the changes to `.travis.yml` then. Because it seems you wanted to investigate other options.

jens-mueller-sociomantic:
Rebased.

joseph-wakeling-sociomantic:
Yes, I'm fine as a first measure taking the `.travis.yml` changes.  So please drop the `[WIP]` from the commit title ;-)

For the future we could look at what is done e.g. for dhtnode:
https://github.com/sociomantic-tsunami/dhtnode/blob/master/.travis.yml

... where job templates are used to run a variety of things for each build.  But in the short term it doesn't seem worth the hassle when we have a test solution that is sufficient (albeit verbose in terms of numbers of builds).

jens-mueller-sociomantic:
Updated the commit message.

joseph-wakeling-sociomantic:
LGTM in principle, but I note that the CI runs don't show any explicit info from the unittests or integration tests as to which engine is running.  I assume this is `stderr` output which is not being tracked by the CI log?

joseph-wakeling-sociomantic:
I'm thinking of this kind of output, which I get when running tests locally:

```
[12:43:56] src/engine/engine.cc:55: MXNet start using engine: NaiveEngine
```

jens-mueller-sociomantic:
I'll check. I find it strange that Travis would not track stderr.

joseph-wakeling-sociomantic:
AFAICS this output is generated using dmlc's logging functionality, which appears to be using the google logging library under the hood.  So it might be about some environment variable which determines the log level.

joseph-wakeling-sociomantic:
Let's see if we can track this down.  Otherwise LGTM.

joseph-wakeling-sociomantic:
See: http://rpg.ifi.uzh.ch/docs/glog.html for some info.  I think that by default it logs to stdout as there is a `--logtostderr=1` option.

However, it may be that, without the google gflags library installed, we need to manually set some environment variables in order to get the wanted output.  See the link, in the section `Setting flags`.

joseph-wakeling-sociomantic:
I think it might work to set the environment variable:

```
GLOG_stderrthreshold=0
```

... as according to the posted link, this should determine the log level above which messages will be forwarded to `stderr`.  (0 means `INFO` level.)

joseph-wakeling-sociomantic:
(I've just been cheeky and pushed a patch with this to your branch, just to see if it works.)

jens-mueller-sociomantic:
Do you have the google logging library installed locally? Because I don't but it works for me locally.

joseph-wakeling-sociomantic:
No.  I do have `libgflags` installed though.  But the patch I pushed makes no difference, it seems.

joseph-wakeling-sociomantic:
OK, here's an interesting thing: I see the log message when building the v0.x.x branch locally but not when building this PR branch.  So I think the environment variable may not be propagating adequately to the tests.

joseph-wakeling-sociomantic:
Compare to e.g. https://travis-ci.org/sociomantic-tsunami/dmxnet/jobs/391675687 for the most recent v0.x.x CI build, where the log message shows up.

joseph-wakeling-sociomantic:
I reckon we need to tweak `BEAVER_DOCKER_VARS`.  Will push a patch with that.

jens-mueller-sociomantic:
I had the same idea but then I thought it only applies to `beaver run`. Maybe it is for all `beaver` commands.

joseph-wakeling-sociomantic:
Out of curiosity, do you know what MXNet does if the environment variable is not explicitly specified?  Does it have some rule for selecting the best choice of engine?

joseph-wakeling-sociomantic:
The `BEAVER_DOCKER_VARS` tweak seems to work, anyway.  Feel free to squash these tweaks into your patches as seems best.

jens-mueller-sociomantic:
It defaults to the `ThreadedEnginePerDevice` (see https://github.com/apache/incubator-mxnet/blob/master/src/engine/engine.cc#L32-L58).

joseph-wakeling-sociomantic:
In that case I would suggest squashing together the last 4 commits from this branch (i.e. my 2 tweaks plus your 2 last patches) into one big `Run jobs for all engines on Travis` patch.

What you could also do is to tweak `Config.mak` so that if it is not already set, the var will be set to `NaiveEngine`.  I think:

```mak
MXNET_ENGINE_TYPE ?= NaiveEngine
```

... will do.  That's probably useful for local testing because it ensures that local calculations will always be performed properly and not run into multithreading issues.

jens-mueller-sociomantic:
Apparently one needs to set them also for the `dlang make` command (see https://github.com/sociomantic-tsunami/beaver/blob/v0.6.x/bin/dlang/make#L19).

jens-mueller-sociomantic:
Hmm. Strange that it segfaults only for D2 with the `ThreadedEngine`. And only `production`.

joseph-wakeling-sociomantic:
Yea, it's worrying.  Anyway, want to fix up and take from here?  We need to find out if the problem is consistent.

joseph-wakeling-sociomantic:
I don't get this segfault locally, FWIW.

jens-mueller-sociomantic:
What do we do if it is not consistent? Because it worked now.

joseph-wakeling-sociomantic:
Did you re-run the build?  It looks like we've lost access to the log of the build that failed :-\

joseph-wakeling-sociomantic:
I wanted to confirm the exact circumstances in which it occurred, just to check if it happened when building or running the unittests.

jens-mueller-sociomantic:
Yes, I thought it would keep the log and just do another build.

joseph-wakeling-sociomantic:
Well, let's squash things down to a mergable set of commits, and take it from there.  If it shows up again let's examine carefully when/how it occurs.

Bear in mind that it might be an upstream bug and not on our side.

jens-mueller-sociomantic:
I triggered it five times on Travis. Didn't show up so far.

joseph-wakeling-sociomantic (approved):
Let's take this in and move forward from here.

PR #55, opened 2018-02-01
Pass DMD environment variable to be available to beaver

jens-mueller-sociomantic:
Even though the `DMD` environment variable was set inside Travis it
seems it was never passed to beaver. Meaning we never performed D2
builds. This change fixes it by passing the `DMD` environment variable
to beaver.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
-        - BEAVER_DOCKER_VARS="F"
+        - BEAVER_DOCKER_VARS="F DMD"
     matrix:
         - DMD=dmd1 DIST=xenial F=production
         - DMD=dmd1 DIST=xenial F=devel


jens-mueller-sociomantic:
Not sure this is actually needed. It seems it was passed. There seem to be some variables that are passed by default.

jens-mueller-sociomantic:
It turns out [these `makd` variables ](https://github.com/sociomantic-tsunami/beaver/blob/v0.x.x/bin/dlang/make#L19) are passed by default. This means we can stop passing `F` explicitly.

jens-mueller-sociomantic:
I suggest closing this and providing another PR to clean up `F` from `BEAVER_DOCKER_VARS` in `.travis.yml`.

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic can I ask you to provide review and recommendations here?

This is a slightly freaky discovery ... at least I was making sure to do D2 test builds locally :-P

jens-mueller-sociomantic:
Not needed anymore. `DMD` is passed default. `F` now as well.

joseph-wakeling-sociomantic:
Ah, this makes sense.  I remember there was some mess-up in the past over what was passed by default and what wasn't, but we had to work around it in the short term.  Let's close and clean up `.travis.yml` separately as you suggest.

PR #56, opened 2018-02-01
Drop explicit passing of environment variable `F` to beaver

jens-mueller-sociomantic:
The environment variable `F` is included implicitly in the variables
passed to beaver via `BEAVER_DOCKER_VARS`. This change removes passing
it explicitly. The then empty environment variable is also removed.


.travis.yml
```diff
@@ -23,7 +23,7 @@ env:
     global:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
-        - BEAVER_DOCKER_VARS="F"
+        - BEAVER_DOCKER_VARS=""
```
joseph-wakeling-sociomantic: Do we even need to assign this variable at all if it's to be empty?  Why not just drop the line?

jens-mueller-sociomantic: I'd like to keep it. Just as a reminder that you may need to pass additional variables. I mean I can also remove it here, but then it will be added for testing with different MXNet engines again.

joseph-wakeling-sociomantic: I'm not sure we need to pass the engines via travis CI.  TBH I would rather have a guarantee that integration tests _always_ test with every single engine (so you can't forget to do this when running integration tests locally).

jens-mueller-sociomantic: You mean we add it somehow to the `Build.mak`. That is also good. But local testing time may then be three times as long.

joseph-wakeling-sociomantic: For the integration tests, that means about 6 seconds instead of 1.5 :-P  I take the point about unittests, though.  Maybe your approach is preferable.

What's bothering me is that we have this combinatorical explosion of options for the build.  It'd be nice to be able to automate that instead of having to manually write out all the combinations.

jens-mueller-sociomantic: I looked whether Travis has some settings to only name the variables and let it implicitly define all combinations. This didn't work.

leandro-lucarella-sociomantic (dismissed):
Normally I would also remove the empty variable, but is up to you!

joseph-wakeling-sociomantic:
Honestly, I'd rather just remove the line if it's not needed ... ?

jens-mueller-sociomantic:
Alright. I remove it then. When needed it will be introduced.

jens-mueller-sociomantic:
Removed the empty environment variable.

jens-mueller-sociomantic:
Rebased.

joseph-wakeling-sociomantic (approved):

PR #60, opened 2018-05-17
[neptune] Update submodules

jenkins-sociomantic:
This PR has been automatically created by *neptune-autopr* and
updates the following submodules:

* **makd** from version **v2.1.3** to version **v2.2.0** ([Release notes](https://github.com/sociomantic-tsunami/makd/releases/tag/v2.2.0), [Changed commits](https://github.com/sociomantic-tsunami/makd/compare/v2.1.3...v2.2.0))
* **ocean** from version **v3.5.11** to version **v3.5.12** ([Release notes](https://github.com/sociomantic-tsunami/ocean/releases/tag/v3.5.12), [Changed commits](https://github.com/sociomantic-tsunami/ocean/compare/v3.5.11...v3.5.12))

You can modify this behavior using your `.neptune.yml` file
as described in the [documentation](https://github.com/sociomantic-tsunami/neptune/blob/master/doc/neptune-metadata.rst#automatic-update-prs).


jens-mueller-sociomantic (dismissed):

joseph-wakeling-sociomantic:
I think we should probably close this PR in favour of a manual one that updates to a much more recent ocean.  We're still in v0.x.y release series, so we can do this without an issue.

joseph-wakeling-sociomantic:
I'll close once such an update is complete.

joseph-wakeling-sociomantic (approved):
I guess this is fine to take in.  I'll do the manual ocean update afterwards.

jens-mueller-sociomantic:
Why didn't we get the beaver update here?

joseph-wakeling-sociomantic:
Good question.

mathias-baumann-sociomantic:
If it was a minor update, you wouldn't get it unless you change your defaults to accept minor updates.

Because this is marked as a library, it receives only patch updates automatically. The motivation behind this is, that we want libraries to have older dependencies so that the applications using them aren't forced to update. Basically, keeping the consequences of updating as low as possible for the things depending on the library

jens-mueller-sociomantic:
We have configured it to receive minor updates for `makd` and `beaver` (see https://github.com/sociomantic-tsunami/dmxnet/blob/v0.x.x/.neptune.yml#L3-L7).

joseph-wakeling-sociomantic:
Right, we'd expect the library default to be respected in general, but not where we explicitly specify an override.

joseph-wakeling-sociomantic:
If there is a hard policy decision here, that's fine, but please let us know.  I would expect per-submodule overrides to be supported (and note that our use case is entirely valid: treating CI differently from actual code dependencies).

mathias-baumann-sociomantic:
Whatever you configured should work. Could you write exactly what you expected to happen and what actually happened? Including the times it should have happened? I can then investigate.

joseph-wakeling-sociomantic:
What we expected to happen was that we should get minor-release patches for makd and beaver (since there are more recent releases than the ones currently used by dmxnet).

The PR to update the autopr settings was merged a week ago: https://github.com/sociomantic-tsunami/dmxnet/pull/62 so we'd expect to have received such PRs on at least one day since then.  But we never got any.

mathias-baumann-sociomantic:
I take it with the latest changes those problems are fixed now (marking beaver as library and makd being rather recent)

joseph-wakeling-sociomantic:
We got an update PR, yes.  Thanks! :-)

PR #62, opened 2018-06-12
Enable automatic update PRs for minor versions for makd and beaver

jens-mueller-sociomantic:
This change enables providing automatic update PRs to minor versions for
all submodules except ocean for which only patch versions should be
provided.

These changes should result in an automatic PR for updating makd to
v2.2.0 and beaver to v0.6.0.

### DIFF
--- a/.neptune.yml
+++ b/.neptune.yml
@@ -1,2 +1,7 @@
 library: true
 d2ready: true
+automatic-update-prs:
+    default: patch
+    override:
+        makd: minor
+        beaver: minor


.neptune.yml
```diff
@@ -1,2 +1,6 @@
 library: true
 d2ready: true
+automatic-update-prs:
+    default: minor
+    override:
+        ocean: patch
```
joseph-wakeling-sociomantic: For a library it is preferable to set `default: patch` and to specify the exceptions for which `minor` is acceptable.

joseph-wakeling-sociomantic: This patch is approaching things with the assumption that ocean is the exception to the rule, but that will cease to be the case the moment we ever accept another D library as a dependency.  It's far preferable to accidentally fail to get minor updates for a CI tool, than it is to accidentally _get_ a minor update for a source library where we don't actually want it.

joseph-wakeling-sociomantic (changes requested):
Please rework this to use the normal `default: patch` as required by libraries.

jens-mueller-sociomantic:
```diff
diff --git a/.neptune.yml b/.neptune.yml
index 3063b68..c5a4c29 100644
--- a/.neptune.yml
+++ b/.neptune.yml
@@ -1,6 +1,7 @@
 library: true
 d2ready: true
 automatic-update-prs:
-    default: minor
+    default: patch
     override:
-        ocean: patch
+        makd: minor
+        beaver: minor
```

joseph-wakeling-sociomantic (approved):
LGTM!  Thanks :-)

PR #64, opened 2018-06-13
Export MXNET_ENGINE_TYPE from Config.mak if not already set

joseph-wakeling-sociomantic:
This should provide a sensible default engine for local testing, which which may readily be overridden by the user, e.g.:

    MXNET_ENGINE_TYPE=ThreadedEngine make test all

It should not interfere with cases where `MXNET_ENGINE_TYPE` is already set in the build environment.

As a friendly by-product, since this tweak ensures `MXNET_ENGINE_TYPE` is always set, libmxnet should now always log the choice when running tests.

### DIFF
--- a/Config.mak
+++ b/Config.mak
@@ -1,5 +1,8 @@
 INTEGRATIONTEST := integrationtest
 
+MXNET_ENGINE_TYPE ?= NaiveEngine
+export MXNET_ENGINE_TYPE
+
 ifeq ($(DVER),2)
 	DC ?= dmd-transitional
 endif


joseph-wakeling-sociomantic:
Damn, the original patch doesn't work: it just means that `NaiveEngine` is used in all cases.  I've pushed a `squash!` patch with a potential fix.

joseph-wakeling-sociomantic:
... and again. My make-fu obviously needs some work :-P

joseph-wakeling-sociomantic:
This should now work, but in preparing it I noticed a bug in the travis setup.  A fix is provided in https://github.com/sociomantic-tsunami/dmxnet/pull/65.

joseph-wakeling-sociomantic:
OK, this appears to have hit the occasional segfault we've observed before: https://api.travis-ci.org/v3/job/391836746/log.txt

```
[00m   [00;33mrun[00m build/production/tmp/allunittests
[00;31mDebug reporting works: true at ./src/mxnet/Handle.d:465
[16:32:52] src/engine/engine.cc:55: MXNet start using engine: ThreadedEngine
[00mSegmentation fault (core dumped)
submodules/makd/Makd.mak:608: recipe for target '/home/travis/build/sociomantic-tsunami/dmxnet/build/production/tmp/allunittests.stamp' failed
make: *** [/home/travis/build/sociomantic-tsunami/dmxnet/build/production/tmp/allunittests.stamp] Error 139
```

This makes me wonder if there's something about the instantiation of `libmxnet` data structures that we are failing to wait for in our unittests.

joseph-wakeling-sociomantic:
(Note that the integration tests run fine, although whether that's down to things working right, or just coincidence, is not obvious.)

jens-mueller-sociomantic:
The segfault is again only D2. Right? Because it seems to occur only in production with the `ThreadedEngine`.

joseph-wakeling-sociomantic:
Have only observed so far for D2, yes.

joseph-wakeling-sociomantic:
Rebased on latest v0.x.x branch.

jens-mueller-sociomantic:
It seems I could reproduce the error locally.

joseph-wakeling-sociomantic:
Consistently?  Does this give you any sense of what's going on (e.g. are you able to reproduce it running in gdb)?

jens-mueller-sociomantic:
Currently just once. The stacktrace is not useful at all.

joseph-wakeling-sociomantic:
I've opened a separate issue for the segfault: https://github.com/sociomantic-tsunami/dmxnet/issues/66.  In the meanwhile, does this PR look OK?

jens-mueller-sociomantic (approved):

PR #71, opened 2018-06-21
Update build scripts to always run tests for all MXNet engine types

joseph-wakeling-sociomantic:
This patch updates `Build.mak` to ensure that unittests and integration tests will always be run for all of the supported engine types.  As well as making it easier to run tests for all engines locally, this will make it possible to greatly simplify our CI setup.

### DIFF
--- a/Build.mak
+++ b/Build.mak
@@ -21,7 +21,7 @@ $(1)-$(2).stamp: $1
 endef
 
 # helper function to generate targets for per-engine test runs
-test_with_engines = $(foreach engine,$2,\
+test_with_engines = $(foreach engine,$(TEST_MXNET_ENGINES),\
 	$(eval $(call run_test_with_engine,$1,$(engine))))
 
 # extra build dependencies for integration tests
@@ -29,7 +29,7 @@ $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
 # run integration tests with all specified engines
-$(eval $(call test_with_engines,$O/test-mxnet,$(TEST_MXNET_ENGINES)))
+$(eval $(call test_with_engines,$O/test-mxnet))
 
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
@@ -39,7 +39,7 @@ $O/test-mxnet.stamp: download-mnist
 $O/%unittests: override LDFLAGS += -lz
 
 # run unittests with all specified engines
-$(eval $(call test_with_engines,$O/allunittests,$(TEST_MXNET_ENGINES)))
+$(eval $(call test_with_engines,$O/allunittests))
 
 $O/allunittests.stamp:
 	$Vtouch $@ # override default implementation
--- a/Build.mak
+++ b/Build.mak
@@ -15,22 +15,34 @@ download-mnist: $C/script/download-mnist
 # helper function to run tests
 run_test = $(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
 
+# helper template to define targets for per-engine test runs
+define run_test_with_engine
+$(1).stamp: $(1)-$(2).stamp
+
+$(1)-$(2).stamp: $1
+	$(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
+endef
+
+# helper function to generate targets for per-engine test runs
+test_with_engines = $(foreach engine,$2,\
+	$(eval $(call run_test_with_engine,$1,$(engine))))
+
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
+# run integration tests with all specified engines
+$(eval $(call test_with_engines,$O/test-mxnet,$(TEST_MXNET_ENGINES)))
+
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
-$O/test-mxnet.stamp: $O/test-mxnet download-mnist
-	$(call run_test,$<,NaiveEngine)
-	$(call run_test,$<,ThreadedEngine)
-	$(call run_test,$<,ThreadedEnginePerDevice)
-	$Vtouch $@
+$O/test-mxnet.stamp: download-mnist
+	$Vtouch $@ # override default implementation
 
 $O/%unittests: override LDFLAGS += -lz
 
-$O/allunittests.stamp: $O/allunittests
-	$(call run_test,$<,NaiveEngine)
-	$(call run_test,$<,ThreadedEngine)
-	$(call run_test,$<,ThreadedEnginePerDevice)
-	$Vtouch $@
+# run unittests with all specified engines
+$(eval $(call test_with_engines,$O/allunittests,$(TEST_MXNET_ENGINES)))
+
+$O/allunittests.stamp:
+	$Vtouch $@ # override default implementation
--- a/Build.mak
+++ b/Build.mak
@@ -12,12 +12,25 @@ endif
 download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
 
+# helper function to run tests
+run_test = $(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
+
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
-$O/test-mxnet.stamp: download-mnist
+$O/test-mxnet.stamp: $O/test-mxnet download-mnist
+	$(call run_test,$<,NaiveEngine)
+	$(call run_test,$<,ThreadedEngine)
+	$(call run_test,$<,ThreadedEnginePerDevice)
+	$Vtouch $@
 
 $O/%unittests: override LDFLAGS += -lz
+
+$O/allunittests.stamp: $O/allunittests
+	$(call run_test,$<,NaiveEngine)
+	$(call run_test,$<,ThreadedEngine)
+	$(call run_test,$<,ThreadedEnginePerDevice)
+	$Vtouch $@
--- a/Build.mak
+++ b/Build.mak
@@ -13,33 +13,49 @@ download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
 
 # helper template to define targets for per-engine test runs
-define run_test_with_engine
+#
+# Params:
+#     $1 = path to executable that runs tests
+#     $2 = name of MXNet engine to use
+define test_with_engine
 $(1).stamp: $(1)-$(2).stamp
 
 $(1)-$(2).stamp: $1
 	$(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
+	$Vtouch $$@
 endef
 
 # helper function to generate targets for per-engine test runs
-test_with_engines = $(foreach engine,$(TEST_MXNET_ENGINES),\
-	$(eval $(call run_test_with_engine,$1,$(engine))))
+#
+# Params:
+#     $1 = path to executable that runs tests
+define run_test_with_engines
+$(foreach engine,$(TEST_MXNET_ENGINES),\
+	$(eval $(call test_with_engine,$1,$(engine))))
+
+$(1).stamp:
+	$Vtouch $$@  # override default implementation
+endef
+
+define run_test_with_dependency
+$(foreach engine,$(TEST_MXNET_ENGINES),\
+	$(eval $(1)-$(engine).stamp: $2))
+endef
 
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
-# run integration tests with all specified engines
-$(eval $(call test_with_engines,$O/test-mxnet))
-
 # extra runtime dependencies for integration tests
-$O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
-$O/test-mxnet.stamp: download-mnist
-	$Vtouch $@ # override default implementation
+$(eval $(call run_test_with_dependency,$O/test-mxnet,\
+	override ITFLAGS += $(MNIST_DATA_DIR)))
+$(eval $(call run_test_with_dependency,$O/test-mxnet,\
+	download-mnist))
+
+# run integration tests with all specified engines
+$(eval $(call run_test_with_engines,$O/test-mxnet))
 
 $O/%unittests: override LDFLAGS += -lz
 
 # run unittests with all specified engines
-$(eval $(call test_with_engines,$O/allunittests))
-
-$O/allunittests.stamp:
-	$Vtouch $@ # override default implementation
+$(eval $(call run_test_with_engines,$O/allunittests))
--- a/Build.mak
+++ b/Build.mak
@@ -12,9 +12,6 @@ endif
 download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
 
-# helper function to run tests
-run_test = $(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
-
 # helper template to define targets for per-engine test runs
 define run_test_with_engine
 $(1).stamp: $(1)-$(2).stamp
--- a/Config.mak
+++ b/Config.mak
@@ -1,5 +1,7 @@
 INTEGRATIONTEST := integrationtest
 
+TEST_MXNET_ENGINES ?= NaiveEngine ThreadedEngine ThreadedEnginePerDevice
+
 MXNET_ENGINE_TYPE ?= NaiveEngine
 export MXNET_ENGINE_TYPE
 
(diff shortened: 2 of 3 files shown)


joseph-wakeling-sociomantic:
I'm going to follow up with a separate travis config patch.  For now I want to see the behavioural difference without it ;-)

Build.mak
```diff
@@ -12,12 +12,25 @@ endif
...
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
-$O/test-mxnet.stamp: download-mnist
+$O/test-mxnet.stamp: $O/test-mxnet download-mnist
+	$(call run_test,$<,NaiveEngine)
+	$(call run_test,$<,ThreadedEngine)
+	$(call run_test,$<,ThreadedEnginePerDevice)
```
joseph-wakeling-sociomantic: It occurs to me that an even better option would be to have a variable defining the supported engines (set in `Config.mak` if not already defined) and just `foreach` over the entries in that variable.  But I'm finding it tricky to work out how best to do that :-\

joseph-wakeling-sociomantic: Specifically: how to implement an appropriate for loop.  Although I guess a simple, manual way is to have a separate `run_test_with_all_engines` function which takes a single parameter (the test to run).

joseph-wakeling-sociomantic:
I've pushed a follow-up patch to simplify the Travis config.

Build.mak
joseph-wakeling-sociomantic: I've tried with:

```
run_test_with_all_engines = $(foreach engine,$(SUPPORTED_MXNET_ENGINES),$(call run_test,$1,$(engine)))
```

... but for some reason, this only ever picks up on the first entry in the `SUPPORTED_MXNET_ENGINES` variable.

leandro-lucarella-sociomantic (dismissed):
LGTM, although I suggest to still provide the option to run one individual test, in case only one fails, so you don't have to repeat all of them.

leandro-lucarella-sociomantic:
(I would also squash)

joseph-wakeling-sociomantic:
> LGTM, although I suggest to still provide the option to run one individual test, in case only one fails, so you don't have to repeat all of them.

Agreed -- it's what I was trying to do by defining a variable defining the engines to test.  Do you have any suggestions for how to do the kind of foreach loop I had in mind above?

Build.mak
leandro-lucarella-sociomantic: BTW, the dependency of `$O/test-mxnet` is added automatically by Makd, you don't need to put it explicitly.

joseph-wakeling-sociomantic: This is what I assumed, but it didn't work without it. :-\

joseph-wakeling-sociomantic:
I've pushed a `squash!` patch that I think gets things working w.r.t. allowing the engines to be specified via an environment variable.  @mihails-strasuns-sociomantic @leandro-lucarella-sociomantic what do you think?  I suspect it's probably possible to simplify.

joseph-wakeling-sociomantic:
... and one more minor tweak.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
+# run integration tests with all specified engines
+$(eval $(call test_with_engines,$O/test-mxnet,$(TEST_MXNET_ENGINES)))
+
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
 $O/test-mxnet.stamp: download-mnist
+	$Vtouch $@ # override default implementation
```
joseph-wakeling-sociomantic: This tweak (and the similar one for unittests) is necessary to avoid the "normal" integration test run from also taking place.  Would be nice to avoid if possible though.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
+# helper function to generate targets for per-engine test runs
+test_with_engines = $(foreach engine,$2,\
+	$(eval $(call run_test_with_engine,$1,$(engine))))
+
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
+# run integration tests with all specified engines
+$(eval $(call test_with_engines,$O/test-mxnet,$(TEST_MXNET_ENGINES)))
```
joseph-wakeling-sociomantic: It'd be nice to be able to simplify these lines.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
+# helper template to define targets for per-engine test runs
+define run_test_with_engine
+$(1).stamp: $(1)-$(2).stamp
+
+$(1)-$(2).stamp: $1
+	$(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
+endef
+
+# helper function to generate targets for per-engine test runs
+test_with_engines = $(foreach engine,$2,\
```
joseph-wakeling-sociomantic: Probably here we can drop `$2` and just use `$(TEST_MXNET_ENGINES)` directly.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
+# helper function to generate targets for per-engine test runs
+test_with_engines = $(foreach engine,$(TEST_MXNET_ENGINES),\
+	$(eval $(call run_test_with_engine,$1,$(engine))))
+
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
+# run integration tests with all specified engines
+$(eval $(call test_with_engines,$O/test-mxnet))
```
joseph-wakeling-sociomantic: Is there any way to avoid the `$(eval $(call ...` here?   They both seem to be necessary, otherwise I get a `missing separator` error.

Config.mak
```diff
@@ -1,5 +1,7 @@
 INTEGRATIONTEST := integrationtest
 
+TEST_MXNET_ENGINES ?= NaiveEngine ThreadedEngine ThreadedEnginePerDevice
```
joseph-wakeling-sociomantic, line 3: One small note here: it should not be possible that the `TEST_MXNET_ENGINES` variable be _empty_.  If I do:

```
TEST_MXNET_ENGINES= make test all
```

... then obviously, nothing runs.  But there should be a failure in this case (which doesn't happen right now).  Any thoughts on the best way to implement this?

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
+define run_test_with_engine
+$(1).stamp: $(1)-$(2).stamp
+
+$(1)-$(2).stamp: $1
+	$(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
+endef
+
+# helper function to generate targets for per-engine test runs
+test_with_engines = $(foreach engine,$(TEST_MXNET_ENGINES),\
+	$(eval $(call run_test_with_engine,$1,$(engine))))
```
joseph-wakeling-sociomantic: Both `eval` and `call` seem necessary here too to avoid a `missing separator` error.

Config.mak
jens-mueller-sociomantic: I don't see why there should be a failure. Because with empty `TEST_MXNET_ENGINES` you define no additional targets.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
 download-mnist: $C/script/download-mnist
 	$(call exec,sh $(if $V,,-x) $^,$(MNIST_DATA_DIR),$^)
 
+# helper template to define targets for per-engine test runs
+define run_test_with_engine
+$(1).stamp: $(1)-$(2).stamp
+
+$(1)-$(2).stamp: $1
+	$(call exec,MXNET_ENGINE_TYPE=$2 $1,$1,$2)
```
jens-mueller-sociomantic: Don't you miss a `$Vtouch $(1)-$(2).stamp` here?

Build.mak
jens-mueller-sociomantic: I don't think so. `call` is needed to run the function and `eval` is needed to evaluate the returned string describing the target (https://www.gnu.org/software/make/manual/html_node/Eval-Function.html#Eval-Function).

Build.mak
jens-mueller-sociomantic: Another approach is to have the MXNet default engine always executed and then change `TEST_MXNET_ENGINES` to mean additional engines to test.

Build.mak
joseph-wakeling-sociomantic: Yes, most likely.  I think I tried using `$@` and there was some objection to that (which I assume is about ordering of make's expansion of these tokens).

Config.mak
joseph-wakeling-sociomantic: @jens-mueller-sociomantic right now they are not ADDITIONAL targets.  Of course, we could make it so.  It just means that `NaiveEngine` tests will run twice.  But if we kill the `MXNET_ENGINE_TYPE ?=` line in `Config.mak`, maybe that has sense (as we then test the case where `MXNET_ENGINE_TYPE` is not specified in the environment).

Build.mak
joseph-wakeling-sociomantic: Yes.  It would probably be best in that case to remove the `MXNET_ENGINE_TYPE ?=` line from `Config.mak` as then we test the default case where the engine is not specified in the environment.

joseph-wakeling-sociomantic: That would also avoid the need to override the default environment with lines like

```
$O/allunittests.stamp:
    $Vtouch $@
```

... so it's probably better this way.

Build.mak
```diff
@@ -12,12 +12,34 @@ endif
...
+
 # extra runtime dependencies for integration tests
 $O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
 $O/test-mxnet.stamp: download-mnist
+	$Vtouch $@ # override default implementation
 
 $O/%unittests: override LDFLAGS += -lz
+
+# run unittests with all specified engines
+$(eval $(call test_with_engines,$O/allunittests))
```
joseph-wakeling-sociomantic: Minor note: would it be worth killing the need to write `$O` here?  i.e. have the calls be

```
$(eval $(call test_with_engines,allunittests)
$(eval $(call test_with_engines,test-mxnet)
```

... and do the `$O` stuff under the hood ... ?

Build.mak
joseph-wakeling-sociomantic: HOWEVER, note that this then messes up things from another point of view, which is: what if I want to specify just _one_ engine to test with, not as an extra?

joseph-wakeling-sociomantic: Obviously I can do `MXNET_ENGINE_TYPE=MyEngine TEST_MXNET_ENGINES= make test all` but this is getting finnicky.  That's why from a CI perspective I'd rather have just one variable (the build script's variable) be the one the user has to care about.

joseph-wakeling-sociomantic:
I've pushed a further fixup patch.  @leandro-lucarella-sociomantic @nemanja-boric-sociomantic you may want to review this solution ;-)

Build.mak
```diff
@@ -12,12 +12,50 @@ endif
...
+
 # extra build dependencies for integration tests
 $O/test-mxnet: override LDFLAGS += -lz
 $O/test-mxnet: override DFLAGS += -debug=MXNetHandleManualFree
 
 # extra runtime dependencies for integration tests
-$O/test-mxnet.stamp: override ITFLAGS += $(MNIST_DATA_DIR)
-$O/test-mxnet.stamp: download-mnist
+$(eval $(call run_test_with_dependency,$O/test-mxnet,\
+	override ITFLAGS += $(MNIST_DATA_DIR)))
```
joseph-wakeling-sociomantic, line 51: I'm not convinced that this is working.  The output of `make -p` gives e.g.:

```
build/devel/tmp/test-mxnet-NaiveEngine.stamp: ITFLAGS += ~/data/mnist
```

(where `~` is expanded;-) but the integration tests (and download script) both use the default value of this parameter, despite it being set in my `Config.local.mak`.

That said, this might be something other than how this `Build.mak` is written.

Build.mak
```diff
@@ -12,12 +12,50 @@ endif
...
+$(eval $(call run_test_with_dependency,$O/test-mxnet,\
+	download-mnist))
+
+# run integration tests with all specified engines
+$(eval $(call run_test_with_engines,$O/test-mxnet))
```
joseph-wakeling-sociomantic, line 56: A weird behaviour: right now `make test` always re-runs the integration tests, despite the associated `.stamp` files being correctly generated.  Any idea what could be responsible?

Build.mak
```diff
@@ -12,12 +12,50 @@ endif
...
 
 $O/%unittests: override LDFLAGS += -lz
+
+# run unittests with all specified engines
+$(eval $(call run_test_with_engines,$O/allunittests))
```
joseph-wakeling-sociomantic, line 61: OTOH this works just fine, with no re-running of unittests after they have been successfully run once.

Build.mak
joseph-wakeling-sociomantic: OK, it looks like the `download-mnist` phony target is responsible, since there is no associated `.stamp`.  It's independent of this PR, just more noticeable.

Build.mak
joseph-wakeling-sociomantic: I've just validated with current v0.x.x, where things are working w.r.t. running `test-mxnet`, but not in terms of where the download script places things (which appears to be an existing `Build.mak` bug).  So there seems to be some problem with how this override is being handled.

joseph-wakeling-sociomantic: It looks like it's because the custom rule implemented in `test_with_engine` is missing the details used in the default makd integration test run:

```
# General rule to Run the test suite binaries
$O/test-%.stamp: $O/test-%
    $(call exec,$< $(_test_drt_opts) $(ITFLAGS),$<,run)
    $Vtouch $@
```

joseph-wakeling-sociomantic: Now, I can easily add `$$(_test_drt_opts) $$(ITFLAGS)` to the call implemented here in `Build.mak`, but note that this is supposed to work for unittests as well as integration tests and there are no similar settings in vanilla makd where unittests are concerned.

@leandro-lucarella-sociomantic @nemanja-boric-sociomantic any advice/thoughts?

joseph-wakeling-sociomantic: `_test_drt_opts` are common, I think, but unittests versus integration-tests should use `UTFLAGS` versus `ITFLAGS` respectively.  I guess I could define a function to work out (based on the full target name) which options should be used.  But it really feels like this is starting to turn into a nasty code duplication between how this `Build.mak` is doing things, versus how makd is doing things.

joseph-wakeling-sociomantic:
Considering how many successive issues we're uncovering, I'm starting to question whether this PR is worth the complications it introduces.

I wonder whether it might be worth trying to distill the essence of what is wanted here (i.e. ability to replay both unittests and integration tests with multiple different runtime environment settings), and implement support directly in makd, to ensure no conflicts or duplication of effort.

Config.mak
nemanja-boric-sociomantic: Perhaps define the `TEST_MXNET_ENGINES` variable to some "always fails" phony target by default, requiring that user override it.

Build.mak
nemanja-boric-sociomantic: TBH, my feeling is you have reached the level of the complexity where we should consider supporting this as a makd feature. @leandro-lucarella-sociomantic

Config.mak
joseph-wakeling-sociomantic: @nemanja-boric-sociomantic I'm not sure it really fixes the issue, because the problem is precisely if the user overrides the variable to set it to empty (a phony target won't solve that).

nemanja-boric-sociomantic: Right, I was talking nonsense. Hm...

nemanja-boric-sociomantic: There's a way of defining default to be `NativeEngine` and filtering it out from the `TEST_MXNET_ENGINES`.

joseph-wakeling-sociomantic: TBH I think this is overthinking things.  If someone sets the field to empty that's their problem ...

Build.mak
joseph-wakeling-sociomantic: Yes, I agree, this looks like it needs makd support.

joseph-wakeling-sociomantic: The question is what level of makd support, though.  It's not clear to me that it needs to be at the level of actually supporting the feature needed here (i.e. running tests multiple times for different environment settings).  It might suffice to provide make functions to run the tests associated with a particular `.stamp` target (so that create a custom recipe without having to do a complete re-duplication of what makd already does).

@leandro-lucarella-sociomantic any thoughts?  (Including "Drop this for now and get on with more actionable items" :-P)

joseph-wakeling-sociomantic:
Removing the milestone and lowering the priority, as I think we should take a bit of a step back from this: so far it's proving more complicated than it's worth for the benefit.  Would still be interested in addressing the slightly bigger picture feature properly at some point, though.

leandro-lucarella-sociomantic:
> I've pushed a further fixup patch. @leandro-lucarella-sociomantic @nemanja-boric-sociomantic you may want to review this solution ;-)

I saw @nemanja-boric-sociomantic already reviewed this, I will skip it then . If you still need my input pplease mention me again!

Build.mak
nemanja-boric-sociomantic: @leandro-lucarella-sociomantic your input is needed here :-)

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic it would be good to have your input on the outcome of our discussions, in particular w.r.t. our discussion point here: https://github.com/sociomantic-tsunami/dmxnet/pull/71#discussion_r197793428

... but I don't think you need to approach this with any urgency.  I can also file an upstream makd issue distilling the requirements (which could be addressed either with a concrete how-to-do-it-downstream solution, or by makd features, or a bit of both, depending on what's required).

leandro-lucarella-sociomantic:
To be honest the discussion is too long and it would take a lot of time to go through it, so if it is an option to write a more generic and summarized issue in makd, that would be the best option IMHO (if appropriate, i.e. if there is anything wrong with it or a new feature is needed).

Issue #66, opened 2018-06-14
Unittests can segfault when using ThreadedEngine


joseph-wakeling-sociomantic:
We have now observed repeated (but flaky) segfaults in CI when `ThreadedEngine` is used.  ~~So far this has only been observed for D2 production builds.~~  For example (https://api.travis-ci.org/v3/job/391836746/log.txt):

```
[00m   [00;33mrun[00m build/production/tmp/allunittests
[00;31mDebug reporting works: true at ./src/mxnet/Handle.d:465
[16:32:52] src/engine/engine.cc:55: MXNet start using engine: ThreadedEngine
[00mSegmentation fault (core dumped)
submodules/makd/Makd.mak:608: recipe for target '/home/travis/build/sociomantic-tsunami/dmxnet/build/production/tmp/allunittests.stamp' failed
make: *** [/home/travis/build/sociomantic-tsunami/dmxnet/build/production/tmp/allunittests.stamp] Error 139
```

joseph-wakeling-sociomantic:
Given the flakiness my suspicion is that there's some sort of race condition in play.  Note that in principle it could be entirely an upstream issue, but this seems less likely because it seems to occur only in the specific case of a D2 production build.

@jens-mueller-sociomantic I would suggest putting `version(none):` at the start of modules successively until the problem stops recurring.  That should at least give some sense of where it derives from.  Then once the module is identified you can selectively `version (none)` individual unittest blocks until you track down the place where the fault occurs.

jens-mueller-sociomantic:
By running the `allunittests` binary in a loop with the `MXNET_ENGINE_TYPE=ThreadedEngine` you can produce the issue consistently. It segfaults after running the tests.

```
dmxnet/build/production/tmp/allunittests  -v -s  -p mxnet. -p integrationtest.
allunittests: unit tests started
allunittests: mxnet.Handle: testing ...Debug reporting works: true at ./src/mxnet/Handle.d:465
 PASS [168us, 4736 bytes (123568 -> 128304)]
allunittests: mxnet.Atomic: testing ... PASS [2us, 0 bytes (127680 -> 127680)]
allunittests: mxnet.Executor: testing ...[11:07:58] src/engine/engine.cc:55: MXNet start using engine: ThreadedEngine
 PASS [2.4ms, 1824 bytes (127680 -> 129504)]
allunittests: mxnet.NDArray: testing ... PASS [28.3ms, 68912 bytes (127680 -> 196592)]
[...]

PR #72, opened 2018-06-22
Shutdown MXNet engine after running unittests

jens-mueller-sociomantic:
This change notifies MXNet to shutdown in the module destructor of
`mxnet.Handle`. This reduces the likelihood of crashes due to a race
happening after executing unittests. The race is likely to be caused by
dmxnet interacting with MXNet while MXNet is cleaning up at process exit
(see sociomantic-tsunami/dmxnet#66 for more details).

Note, that it does not seem to eliminate those crashes. It only seems to
reduce it because there still seems to be the possibility of a race
happening until MXNet actually shuts down. In so far, this is only a
measure to reduce the number of these crashes since after all they
happen at process exit anyway. In future, one should look into making
dmxnet always exit properly when running the unittests.

Part of sociomantic-tsunami/dmxnet#66.

### DIFF
--- a/src/mxnet/API.d
+++ b/src/mxnet/API.d
@@ -29,6 +29,29 @@ version(UnitTest)
     import ocean.core.Test;
 }
 
+version (UnitTest)
+{
+    /***************************************************************************
+
+        Notifies the MXNet engine to shutdown after leaving `main` for
+        unittests
+
+        Shutting down the engine limits a race when interacting with MXNet
+        during process exit. The race happens when MXNet is cleaning up (on
+        process exit).
+
+        Note:
+           This only reduces the probability of the race but it does not seem
+           to eliminate them.
+
+    ***************************************************************************/
+
+    static ~this ()
+    {
+        notifyShutdown();
+    }
+}
+
 
 /*******************************************************************************
 
--- a/src/mxnet/API.d
+++ b/src/mxnet/API.d
@@ -119,3 +119,18 @@ unittest
     invoke!(mxNetSucceeds)(&answer);
     test!("==")(answer, 42);
 }
+
+
+/*******************************************************************************
+
+    Notifies MXNet to shutdown
+
+    By calling this function MXNet is notified to shut down its engine. When
+    the engine is shut down MXNet will not execute pending calculations.
+
+*******************************************************************************/
+
+public void notifyShutdown ()
+{
+    invoke!(MXNotifyShutdown)();
+}


src/mxnet/Handle.d
```diff
@@ -29,6 +30,19 @@ version (UnitTest)
...
 
+// shutting down the engine to limit races when interacting with MXNet during
+// process exit (races happen when MXNet shuts down while making calls to API)
+// this only reduces the probability of races but does not seem to eliminate
+// them
+version(UnitTest)
+{
+    static ~this ()
+    {
+        notifyShutdown();
```
joseph-wakeling-sociomantic: Is there a reason for putting it here instead of `mxnet.API`?  I'm not sure that my question in https://github.com/sociomantic-tsunami/dmxnet/issues/66#issuecomment-399384876 is really sufficient grounds to place it in `mxnet.Handle`.

src/mxnet/Handle.d
```diff
@@ -29,6 +30,19 @@ version (UnitTest)
 }
 
 
+// shutting down the engine to limit races when interacting with MXNet during
+// process exit (races happen when MXNet shuts down while making calls to API)
+// this only reduces the probability of races but does not seem to eliminate
+// them
```
joseph-wakeling-sociomantic: This could be module destructor doc.

src/mxnet/Handle.d
```diff
@@ -29,6 +30,19 @@ version (UnitTest)
...
+version(UnitTest)
```
joseph-wakeling-sociomantic: Minor style: `version (UnitTest)` (space after `version`)

joseph-wakeling-sociomantic (changes requested):
Looks all right, just a few small tweaks/queries.

src/mxnet/Handle.d
jens-mueller-sociomantic: We can place it elsewhere. I think there are only organizational reason. One could put it into `mxnet.API` because after all the problem seems to be with using the MXNet API shuts down. On the other hand the problem seems to be specifically with handles which suggests to put it into `mxnet.Handle`. Either way is fine with me.

src/mxnet/Handle.d
jens-mueller-sociomantic: I think I would make it more public if it fixed the issue. Currently it's for me more like internal `FIXME` kind of doc. But I can change.

src/mxnet/Handle.d
joseph-wakeling-sociomantic: Let's keep it simple in terms of code changes and put it in `mxnet.API` for the time being.  The only reason for doing otherwise would be if it made a difference to the segfault occurring (and IIUC while it greatly reduces it, it doesn't entirely eliminate the segfaulting, no ... ?)

src/mxnet/Handle.d
joseph-wakeling-sociomantic: Fair enough.  Let's leave with just a comment then.

jens-mueller-sociomantic: Hmm. Turned it into module documentation. It doesn't matter much. But I thought it also doesn't hurt to tell users about it.

src/mxnet/API.d
```diff
@@ -32,7 +32,7 @@ version(UnitTest)
 
 /*******************************************************************************
 
-    Notifies the MXNet engine to shutdown after leaving `main`
+    Notifies the MXNet engine to shutdown after leaving `main` for unittests
```
joseph-wakeling-sociomantic: Minor: if you're going to ddoc this I'd suggest putting the ddoc inside the `version (UnitTest)` block.

src/mxnet/API.d
```diff
@@ -30,6 +30,27 @@ version(UnitTest)
...
 
+/*******************************************************************************
+
+    Notifies the MXNet engine to shutdown after leaving `main` for unittests
+
+    Shutting down the engine limits a race when interacting with MXNet during
+    process exit. The race happens when MXNet is cleaning up (on process exit).
+
+    NB This only reduces the probability of the race but it does not seem to
+       eliminate them.
```
joseph-wakeling-sociomantic: Indentation inconsistency here.

jens-mueller-sociomantic: It's on purpose to align with the `NB`. I think it makes it stick out more. Meaning one won't miss it.

joseph-wakeling-sociomantic: You could always do:

```
Note:
    All the stuff you want noted
```

... which will highlight things in the generated ddoc too ;-)

joseph-wakeling-sociomantic (commented):
This LGTM, just needs to be squashed.  (For the record with a small PR like this I think it's better to just make the tweaks in-place and post the diffs in comments, because then we don't have to worry about squashing once it's in a good state.)

joseph-wakeling-sociomantic (approved):

joseph-wakeling-sociomantic:
LGTM, we can take in once CI passes.

PR #73, opened 2018-06-22
Advance ocean from v3.5.12 to v4.2.1

joseph-wakeling-sociomantic:
ocean v3.5.12(d49e328e)...v4.2.1(adaa0114)
    
Release notes: https://github.com/sociomantic-tsunami/ocean/releases/tag/v4.2.1
Changed commits: https://github.com/sociomantic-tsunami/ocean/compare/v3.5.12...v4.2.1

This brings dmxnet's ocean dependency up to date with the latest major release series.  This is a breaking change, but a tolerable one given that we are still in the v0.x.y release series.


jens-mueller-sociomantic:
Just to verify: This requires downstream users to be on ocean v4.1.6 which alright I'll guess.

joseph-wakeling-sociomantic:
Yes.  Note that semver explicitly allows us to make this sort of breaking change in a minor release from the v0.x.y series.

joseph-wakeling-sociomantic:
Updated to the just-release v4.2.0.

jens-mueller-sociomantic (approved):

jens-mueller-sociomantic (changes requested):
There is a `v4.2.1` out now which should be used instead I think.

joseph-wakeling-sociomantic:
Updated.

joseph-wakeling-sociomantic:
BTW, normally, if something like this was passing, I'd suggest just taking it in and letting neptune-autopr take care of the patch releases (we might as well rely on automation for what we can).  But since it looks like there was a CI snafu anyway, might as well update ;-)

jens-mueller-sociomantic (approved):

PR #75, opened 2018-06-25
Simplify docker image build script

joseph-wakeling-sociomantic:
We no longer need to download the sociomantic-tsunami APT key as this should already be in the base image.  We also tweak the filename used for the MXNet repo details, just to keep it cleanly separated from the default tsunami repos.

### DIFF
--- a/docker/build
+++ b/docker/build
@@ -1,18 +1,14 @@
 #!/bin/sh
 set -eux
 
-# Update essential packages
-apt-get update
-apt-get -y upgrade
-
-# Add sociomantic-tsunami APT repos
-apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61
-echo '# sociomantic-tsunami repos' \
-    > /etc/apt/sources.list.d/sociomantic-tsunami.list
+# Add sociomantic-tsunami mxnet APT repo
 dist="$(lsb_release -cs)"
 echo "deb https://dl.bintray.com/sociomantic-tsunami/mxnet $dist release" \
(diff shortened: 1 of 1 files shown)


docker/build
```diff
@@ -5,13 +5,10 @@ set -eux
...
 
-# Add sociomantic-tsunami APT repos
-apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 379CE192D401AB61
-echo '# sociomantic-tsunami repos' \
-    > /etc/apt/sources.list.d/sociomantic-tsunami.list
+# Add sociomantic-tsunami mxnet APT repos
 dist="$(lsb_release -cs)"
 echo "deb https://dl.bintray.com/sociomantic-tsunami/mxnet $dist release" \
-    >> /etc/apt/sources.list.d/sociomantic-tsunami.list
+    >> /etc/apt/sources.list.d/mxnet.list
```
joseph-wakeling-sociomantic: Looking at this I reckon the `apt-get update` and `apt-get -y upgrade` stages could come after the repos are added.

docker/build
```diff
@@ -5,13 +5,10 @@ set -eux
 apt-get update
```
nemanja-boric-sociomantic: I think you can just add the mxnet repo before this and by that you save a single `apt-get update` below?

joseph-wakeling-sociomantic: Indeed.

joseph-wakeling-sociomantic:
Updated both the commit and its description.

nemanja-boric-sociomantic (approved):

PR #76, opened 2018-06-26
Ensure MNIST usage example does not require loading the dataset

joseph-wakeling-sociomantic:
The integration-test MNIST usage example should not require us to really load the dataset, as this makes an implicit assumption that the dataset has already been downloaded.  By using a nested function declaration, we can check the validity of the example code without actually needing to run it.

This should prevent unittest failures on systems where the MNIST dataset has not already been downloaded.

### DIFF
--- a/integrationtest/mxnet/MNIST.d
+++ b/integrationtest/mxnet/MNIST.d
@@ -36,25 +36,31 @@ import ocean.sys.Environment;
 import ocean.transition;
 
 
+// we use a nested function in this unittest example to ensure that
+// the code compiles without actually running it, to avoid the need
+// to really load the dataset
 ///
 unittest
 {
-    // load the MNIST training dataset
-    auto mnist_training = trainingSet(datasetPath([]));
-    auto images = mnist_training.images;
-    auto labels = mnist_training.labels;
+    void useMNIST ()
+    {
+        // load the MNIST training dataset
+        auto mnist_training = trainingSet(datasetPath([]));
+        auto images = mnist_training.images;
+        auto labels = mnist_training.labels;
 
-    // dimensions of an image
-    auto image_rows = 28;
-    auto image_cols = 28;
-    assert(labels.length == images.length / (image_rows * image_cols));
+        // dimensions of an image
+        auto image_rows = 28;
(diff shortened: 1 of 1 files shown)


integrationtest/mxnet/MNIST.d
```diff
@@ -39,22 +39,25 @@ import ocean.transition;
...
-    auto image_cols = 28;
-    assert(labels.length == images.length / (image_rows * image_cols));
-
-    // access the first image and its label
-    auto first_training_image = images[0 .. image_rows * image_cols];
-    auto first_training_label = labels[0];
-
-    // image is stored row wise with values between 0 (white) and 255 (black)
-    auto first_row = first_training_image[0 .. image_rows];
+    void useMNIST ()
```
jens-mueller-sociomantic, line 45: Might be good to add comment here why the function is introduced. Because this is usage example and it might be confusing for a usage example.

joseph-wakeling-sociomantic:
I've added an explanatory comment outside the usage example:

```diff
diff --git a/integrationtest/mxnet/MNIST.d b/integrationtest/mxnet/MNIST.d
index 5a9be6f..89e1eff 100644
--- a/integrationtest/mxnet/MNIST.d
+++ b/integrationtest/mxnet/MNIST.d
@@ -36,6 +36,9 @@ import ocean.sys.Environment;
 import ocean.transition;
 
 
+// we use a nested function in this unittest example to ensure that
+// the code compiles without actually running it, to avoid the need
+// to really load the dataset
 ///
 unittest
 {
```

jens-mueller-sociomantic:
There is also an alternative to exclude unittests from `integrationtest/` via `TEST_FILTER_OUT += $C/integrationtest/%`. You can also only exclude the module `integrationtest.mxnet.MNIST` by using `TEST_FILTER_OUT += $C/integrationtest/mxnet/MNIST.d`.

jens-mueller-sociomantic (approved):

joseph-wakeling-sociomantic:
I'd rather not take that approach because we might want to have other unittests at some point.  Easier to just write the usage example in a way that avoids causing problems.

PR #77, opened 2018-06-28
Supply DIST as part of global CI parameters

joseph-wakeling-sociomantic:
This patch simplifies our CI setup a little: we currently have no need for custom distro jobs, and if we ever do then presumably we can specify a custom `DIST` only for the jobs that need it.


joseph-wakeling-sociomantic:
Rebased on master.

christoph-wasicki-sociomantic (approved):

PR #88, opened 2018-10-19
Convert dmxnet to be D2-only

joseph-wakeling-sociomantic:
This PR applies the D2 conversion changes generated by `make d2conv` and updates the ocean submodule to its corresponding `+d2` release.  The second patch cleans out inappropriate `(&this)` generated by the automated changes, so as to minimize overall PR diff to what matters.

`Build.mak` and `Config.mak` have been updated to drop special-casing for D1 and D2 builds and to hardcode `DVER := 2`, while `.neptune.yml` has been updated to mark the repo as D2-only.

The project `README` has been edited to remove sections on D1 => D2 conversion and to note that the library is D2 only.

Travis CI config has been updated to remove D1 builds and generation of autoconverted `+d2` releases.  Note that a consequence of this is that future D2 releases of dmxnet will have regular vX.y.z version tags, and not vX.y.z+d2.

CI builds have been added for vanilla upstream `dmd` to ensure that the library is not only compatible with `dmd-transitional`.

* submodules/ocean v4.2.10(ed0cb3f3)...v4.2.10+d2(402d4246) (1 commits)
  > Auto-convert to D2


joseph-wakeling-sociomantic:
Hmmm, interesting, `d1to2fix` is still being run.  Looks like there is something missing in setup.

nemanja-boric-sociomantic:
It seems that you're missing `.D2-ready` with `ONLY` as a content:

https://github.com/sociomantic-tsunami/beaver/blob/83d9759688974893916c8b8d6018bca1c1dd8ee4/lib/dlang.sh#L38-L44

joseph-wakeling-sociomantic:
Hah, I wondered if that was needed, but wasn't sure which of our systems cares.  I'll add.

joseph-wakeling-sociomantic:
Updated to set `D2_ONLY=1` in the travis CI environment:

```diff
diff --git a/.travis.yml b/.travis.yml
index 7c33d1e..72503d3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ env:
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
         - BEAVER_DOCKER_VARS="MXNET_ENGINE_TYPE"
         - DIST=xenial
+        - D2_ONLY=1
 
 # Basic config is inherited from the global scope
 jobs:
```

The 1st commit message has been tweaked a bit to reflect this.

joseph-wakeling-sociomantic:
Interesting that this doesn't work, I presume because beaver's `dlang.sh` does not actually check the value of this variable before overriding it based on the presence (or not) of `.D2-ready`.

joseph-wakeling-sociomantic:
Updated again to use `.D2-ready` instead of setting `D2_ONLY` in `.travis.yml`:

```diff
diff --git a/.D2-ready b/.D2-ready
new file mode 100644
index 0000000..bc8c7b4
--- /dev/null
+++ b/.D2-ready
@@ -0,0 +1 @@
+ONLY
diff --git a/.travis.yml b/.travis.yml
index 72503d3..7c33d1e 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ env:
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH)"
         - BEAVER_DOCKER_VARS="MXNET_ENGINE_TYPE"
         - DIST=xenial
-        - D2_ONLY=1
 
 # Basic config is inherited from the global scope
 jobs:
```

joseph-wakeling-sociomantic:
Well, this seems to work.  @nemanja-boric-sociomantic would you like to give a core-team eye on this, since you're here?  No worries if not.

src/mxnet/NDArray.d
```diff
@@ -822,7 +822,7 @@ public alias Tuple!(float, double, int, ubyte) SupportedElementTypes;
 
 public template isSupportedElementType (T)
 {
-    const bool isSupportedElementType =
+    static immutable bool isSupportedElementType =
```
jens-mueller-sociomantic, line 825: I think `enum` would be better here than `static immutable`.

jens-mueller-sociomantic: This could be an `enum`.

src/mxnet/Util.d
```diff
@@ -49,14 +49,14 @@ version(UnitTest)
 
 public mstring toNoLossString (in float x, mstring buf)
 {
-    const max_length = 1 + // optional minus sign
+    static immutable max_length = 1 + // optional minus sign
```
jens-mueller-sociomantic, line 52: This could also be an `enum`. Also realigning the following comments would improve readability.

src/mxnet/Util.d
```diff
@@ -49,14 +49,14 @@ version(UnitTest)
...
                        1 + // leading decimal digit
                        1 + // .
                        8 + // 8 decimal digits
                        1 + // e
                        1 + // sign
                        2;  // decimal exponent
-    const format_string = "{:e8}";
+    static immutable format_string = "{:e8}";
```
jens-mueller-sociomantic, line 59: Could also be an `enum`.

relnotes/d2-only.migration.md
```diff
@@ -0,0 +1,5 @@
+### dmxnet is now a D2-only library
+
+As of this release, dmxnet only supports D2 applications.  Its APIs have
+not changed, but its code has been converted from D1 to D2, and it will
+now only build with a D2 compiler.
```
jens-mueller-sociomantic, line 5: Should we mention here that from now there will be no `+d2` tags anymore. Could be surprising otherwise.

jens-mueller-sociomantic (changes requested):
Mostly asking for some clarification whether it is useful to use `enum`s, alignment of code and the mentioning of dropping `+d2` tags.

src/mxnet/NDArray.d
joseph-wakeling-sociomantic: For now we are not going to change anything in the autoconversion, apart from removing changes that really don't do anything (i.e. the superfluous `(&this)`).  We can do plenty of cleanup afterwards.

BTW there are some reasons to prefer `static immutable` over `enum`.  IIRC it relates to stack size and the number of individual copies that get instantiated.

relnotes/d2-only.migration.md
joseph-wakeling-sociomantic: Yes, sounds reasonable.  I'll update accordingly.

joseph-wakeling-sociomantic:
Updated to add a note on release tags to migration instructions:

```diff
diff --git a/relnotes/d2-only.migration.md b/relnotes/d2-only.migration.md
index 992715c..c2b73ec 100644
--- a/relnotes/d2-only.migration.md
+++ b/relnotes/d2-only.migration.md
@@ -3,3 +3,6 @@
 As of this release, dmxnet only supports D2 applications.  Its APIs have
 not changed, but its code has been converted from D1 to D2, and it will
 now only build with a D2 compiler.
+
+In consequence, D2 releases of dmxnet will now simply be tagged `vX.y.z`
+and not `vX.y.z+d2`.
```

src/mxnet/NDArray.d
jens-mueller-sociomantic: Yes, I was only pointing out this case because here we won't safe any stack size or instantiate less. I tried to only point out the cases where those arguments should not apply. But let's leave this for later.

jens-mueller-sociomantic (dismissed):
LGTM.

joseph-wakeling-sociomantic:
I had to force-push to re-trigger CI hooks; since I needed to do this anyway I took the opportunity to tweak the release notes slightly:

```diff
diff --git a/relnotes/d2-only.migration.md b/relnotes/d2-only.migration.md
index c2b73ec..0542e04 100644
--- a/relnotes/d2-only.migration.md
+++ b/relnotes/d2-only.migration.md
@@ -4,5 +4,5 @@ As of this release, dmxnet only supports D2 applications.  Its APIs have
 not changed, but its code has been converted from D1 to D2, and it will
 now only build with a D2 compiler.
 
-In consequence, D2 releases of dmxnet will now simply be tagged `vX.y.z`
-and not `vX.y.z+d2`.
+In consequence, D2 releases of dmxnet (the only kind from here on out!)
+will now simply be tagged `vX.y.z` and not `vX.y.z+d2`.
```

jens-mueller-sociomantic (approved):

PR #89, opened 2018-10-22
Add closure allocation check to CI stages

joseph-wakeling-sociomantic:
This should allow us to verify that we are not accidentally generating garbage via the use of closures.  We re-use ocean's `closure.sh` script, since its behaviour should be perfectly acceptable for our own use-case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,3 +66,10 @@ jobs:
           env: DMD=dmd-transitional F=devel MXNET_ENGINE_TYPE=ThreadedEngine
         - <<: *test-matrix
           env: DMD=dmd-transitional F=devel MXNET_ENGINE_TYPE=ThreadedEnginePerDevice
+
+        # Additional stages
+
+        - stage: Closure allocation check
+          env: DMD=dmd-transitional F=devel
+          install: beaver dlang install
+          script: ci/closures.sh
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -x -o pipefail
+
+# Enable printing of all potential GC allocation sources to stdout
+export DFLAGS=-vgc
+
+# Passes only if `make` succeeds and `grep` successfully finds no
+# `closure` (exit status 1); fails if either `make` or `grep` fail
+# or if `grep` finds at least one `closure`
+beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)


.travis.yml
```diff
@@ -69,6 +69,11 @@ jobs:
 
         # Additional stages
 
+        - stage: Closure allocation check
+          env: DMD=dmd-transitional DIST=xenial F=devel
```
jens-mueller-sociomantic: Why the `DIST=xenial`?

jens-mueller-sociomantic (commented):
Wondering what the `DIST=xenial` is for.

.travis.yml
joseph-wakeling-sociomantic: Ah yes, this is probably outdated now as `DIST=xenial` is now a global environment variable.

joseph-wakeling-sociomantic:
Updated to remove unnecessary `DIST` setting:

```diff
diff --git a/.travis.yml b/.travis.yml
index 5f51b8b..b48dbca 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ jobs:
         # Additional stages
 
         - stage: Closure allocation check
-          env: DMD=dmd-transitional DIST=xenial F=devel
+          env: DMD=dmd-transitional F=devel
           install: beaver dlang install
           script: submodules/ocean/ci/closures.sh
```

jens-mueller-sociomantic (dismissed):
LGTM.

joseph-wakeling-sociomantic:
Blocking, as I realized that the `closures.sh` script hardcodes a `make d2conv` call.  This will break as soon as we are D2-only.

joseph-wakeling-sociomantic:
Unblocked and updated to support D2-only code.  There was never any need to use the `closures.sh` script: without the need for D2 conversion, we can just invoke `make fasttest` directly and grep for `closure`.

joseph-wakeling-sociomantic:
Updated to add a leading `!` to the command, to ensure we generate an exit status of 0 if _no_ closures are reported, 1 if they are.

```diff
diff --git a/.travis.yml b/.travis.yml
index 2c7f1af..b5bdfc5 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,4 +72,5 @@ jobs:
         - stage: Closure allocation check
           env: DFLAGS=-vgc DMD=dmd-transitional F=devel
           install: beaver dlang install
-          script: beaver dlang make fasttest 2>&1 | grep "closure"
+          # note leading ! to invert result of grep not finding anything
+          script: ! beaver dlang make fasttest 2>&1 | grep "closure"
```

jens-mueller-sociomantic:
Don't we need `DFLAGS=-vgc` then?

jens-mueller-sociomantic:
Ah, you added it in another commit. Because I only looked at the *changes* in the above diff.

joseph-wakeling-sociomantic:
I added it in this commit, but on a different line ;-)

joseph-wakeling-sociomantic:
Looks like we need to use a separate shell script to work around limitations of Travis CI `script:` sections.  I've pushed an example that is designed to fail (it greps for `GC` instead of `closure`), but which should illustrate the principle.

joseph-wakeling-sociomantic:
Updated with what I think should be the final version.  Note that I've moved the `DFLAGS` setting to the shell script, since it doesn't need to exist outside.  This keeps all the closure-specific issues inside the script.

```diff
diff --git a/.travis.yml b/.travis.yml
index 7dcf42d..445fabb 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ jobs:
         # Additional stages
 
         - stage: Closure allocation check
-          env: DFLAGS=-vgc DMD=dmd-transitional F=devel
+          env: DMD=dmd-transitional F=devel
           install: beaver dlang install
           # note leading ! to invert result of grep not finding anything
           script: ci/closures.sh
diff --git a/ci/closures.sh b/ci/closures.sh
index fafb5db..44cbb09 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -1,6 +1,15 @@
 #!/bin/sh
 set -xe
 
+# Enable printing of all potential GC allocation sources to stdout
+export DFLAGS=-vgc
+
+# Run tests and pipe compiler output to temporary file
 tmpfile=$(mktemp /tmp/closure-check.XXXXXXXX)
 beaver dlang make fasttest 2>&1 | tee -a "${tmpfile}"
-! grep -e "GC" "${tmpfile}"
+
+# Ensure there are no lines about closure allocation in
+# the compiler output (note leading ! to invert the grep
+# exit status, since we want to pass if we find nothing,
+# and fail if we find a line containing `closure`)
+! grep -e "closure" "${tmpfile}"
```

joseph-wakeling-sociomantic:
I'm pushing some `squash!` patches temporarily to ensure that we handle the case of `make` failing.  This first one invokes a false target.  Since we do not use `set -o pipefail` the assumption here is that `tee` will swallow up the non-zero exit status (reflecting false-positive CI passes with earlier drafts of this PR).

joseph-wakeling-sociomantic:
Next `squash!` patch adds `-o pipefail`, which ought to ensure that the problematic `make` target results in test failure.

joseph-wakeling-sociomantic:
OK, `-o pipefail` works, now to re-validate the actual `grep` check.  Next patch after this will restore the correct `grep -e "closure"`, and after that we can squash.

joseph-wakeling-sociomantic:
Corrected the `grep` to search for `closure`, but removed a `scope` on a delegate parameter, which should trigger the `closure` search and hence fail the test.

joseph-wakeling-sociomantic:
Replaced the `scope` qualifier.  This should now pass, in which case we can squash and rebase.

joseph-wakeling-sociomantic:
OK, passed!  Will now squash and rebase.

joseph-wakeling-sociomantic:
Rebased and squashed.  I've also added manual cleanup of the temporary file:

```diff
diff --git a/ci/closures.sh b/ci/closures.sh
index 1665274..f274b52 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -13,3 +13,6 @@ beaver dlang make fasttest 2>&1 | tee -a "${tmpfile}"
 # exit status, since we want to pass if we find nothing,
 # and fail if we find a line containing `closure`)
 ! grep -e "closure" "${tmpfile}"
+
+# clean up the temporary file like a good citizen
+rm "${tmpfile}"
```

joseph-wakeling-sociomantic:
Actually, before finalizing, I'd like to see if we can avoid the need for a temporary file; the assumption that it was needed is probably based on a misinterpretation of how the leading `!` works.  I'll follow up with `fixup` patches accordingly.

joseph-wakeling-sociomantic:
OK, it looks like the `!` overrides the overall result, not just the output of the `grep`.  This means that if the `make` command fails, the test still passes!  So I'm going to drop the `squash` patch and return to piping output to a tempfile before grepping as a separate step.

@mihails-strasuns-sociomantic FYI, as it impacts on the correctness of https://github.com/sociomantic-tsunami/ocean/pull/650.

ci/closures.sh
```diff
@@ -4,15 +4,6 @@ set -xe -o pipefail
...
-# the compiler output (note leading ! to invert the grep
-# exit status, since we want to pass if we find nothing,
-# and fail if we find a line containing `closure`)
-! grep -e "closure" "${tmpfile}"
-
-# clean up the temporary file like a good citizen
-rm "${tmpfile}"
+# Run tests and pipe compiler output through `grep` to
+# discover any reported closure GC allocations
+! beaver dlang make fastte 2>&1 | grep -e "closure"
```
leandro-lucarella-sociomantic: `fastte`? :thinking:

joseph-wakeling-sociomantic: Deliberate to see what happened if the `make` command failed.  I was worried about exactly what has happened, which is that the test passed despite the failure.

jens-mueller-sociomantic:
> OK, it looks like the `!` overrides the overall result, not just the output of the `grep`.  This means that if the `make` command fails, the test still passes!  So I'm going to drop the `squash` patch and return to piping output to a tempfile before grepping as a separate step.

Wouldn't this work

```sh
set -o pipefail
beaver dlang make DFLAGS=-vgc fasttest | (! grep "closure")
```

?

joseph-wakeling-sociomantic:
Let's give it a try.

joseph-wakeling-sociomantic:
Squashed, and tweaked comments a bit:

```diff
diff --git a/.travis.yml b/.travis.yml
index 445fabb..e2f3d66 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,5 +72,4 @@ jobs:
         - stage: Closure allocation check
           env: DMD=dmd-transitional F=devel
           install: beaver dlang install
-          # note leading ! to invert result of grep not finding anything
           script: ci/closures.sh
diff --git a/ci/closures.sh b/ci/closures.sh
index 1da6570..455a9e8 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -5,5 +5,7 @@ set -xe -o pipefail
 export DFLAGS=-vgc
 
 # Run tests and pipe compiler output through `grep` to
-# discover any reported closure GC allocations
+# discover any reported closure GC allocations.  Note
+# use of `!` to invert the exit status of `grep`, so
+# that it is 0 if we do _not_ find any GC `closure`.
 beaver dlang make fasttest 2>&1 | (! grep -e "closure")
```

joseph-wakeling-sociomantic:
As a minor note: this is still not entirely adequate as it will "pass" if `grep` itself fails.  But it's better than nothing for now.

joseph-wakeling-sociomantic:
Actually, I think there's a straightforward fix for this.  I'm pushing a few new fixup patches to validate this.

joseph-wakeling-sociomantic:
Damn, how come this is not working?  Presumably because the `grep` command "fails" according to the logic of `set -x` ... ?

jens-mueller-sociomantic:
You mean `set -e` (`set -x` is just tracing, isn't it). But yes that's the problem. grep returns with failure, hence the `test` won't be executed.

joseph-wakeling-sociomantic:
Squashed, and edited the script comment describing the principles behind the check:

```diff
diff --git a/ci/closures.sh b/ci/closures.sh
index 5f8bf65..0298ef6 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -5,7 +5,12 @@ set -x -o pipefail
 export DFLAGS=-vgc
 
 # Run tests and pipe compiler output through `grep` to
-# discover any reported closure GC allocations.  Note
-# use of `!` to invert the exit status of `grep`, so
-# that it is 0 if we do _not_ find any GC `closure`.
+# discover any reported closure GC allocations.
+# Note the explicit check for exit status 1 (meaning that
+# grep did not find `closure`) as opposed to 0 (if grep
+# found `closure`) or 2 (error).
+# Should `make` fail (exit status 2) then `set -o pipefail`
+# will propagate the non-zero exit status, meaning that the
+# script will fail either if `make` fails, if `grep` fails
+# with an error, or if `grep` finds `closure`.
 beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
```

joseph-wakeling-sociomantic:
After thinking it through I've done another rewrite of the script comments to ensure the check is clearly described:

```diff
diff --git a/ci/closures.sh b/ci/closures.sh
index 0298ef6..b528930 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -4,13 +4,15 @@ set -x -o pipefail
 # Enable printing of all potential GC allocation sources to stdout
 export DFLAGS=-vgc
 
-# Run tests and pipe compiler output through `grep` to
-# discover any reported closure GC allocations.
-# Note the explicit check for exit status 1 (meaning that
-# grep did not find `closure`) as opposed to 0 (if grep
-# found `closure`) or 2 (error).
-# Should `make` fail (exit status 2) then `set -o pipefail`
-# will propagate the non-zero exit status, meaning that the
-# script will fail either if `make` fails, if `grep` fails
-# with an error, or if `grep` finds `closure`.
+# Ensure there are no lines about closure allocations in the output.
+# Note explicit check for `grep` exit status 1, i.e. no lines found.
+# This means tests will fail if any of these conditions are met:
+#
+#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
+#   * `grep` fails (exit status 2)
+#   * `grep` finds `closure` (exit status 0)
+#
+# Note that `make` can in principle exit with status 1 too, but only
+# if the `-q` option is used, which should never be the case for our
+# builds, so we can assume exit status 1 only comes from `grep`.
 beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
```

joseph-wakeling-sociomantic:
This is now ready for review.

ci/closures.sh
```diff
@@ -0,0 +1,18 @@
...
+# This means tests will fail if any of these conditions are met:
+#
+#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
+#   * `grep` fails (exit status 2)
+#   * `grep` finds `closure` (exit status 0)
+#
+# Note that `make` can in principle exit with status 1 too, but only
+# if the `-q` option is used, which should never be the case for our
+# builds, so we can assume exit status 1 only comes from `grep`.
+beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
```
jens-mueller-sociomantic, line 10: Why do you do `2>&1`? Also you could remove `-e`.

joseph-wakeling-sociomantic: `2>&1` is just copying what was in the ocean script, and I'm assuming there was a good reason for it.  Without being certain I suspect it's because the compiler output is not guaranteed to go to `stdout`.

joseph-wakeling-sociomantic: I quite like having the explicitness of `-e "closure"`, but YMMV of course.

ci/closures.sh
```diff
@@ -0,0 +1,18 @@
...
+# Note explicit check for `grep` exit status 1, i.e. no lines found.
+# This means tests will fail if any of these conditions are met:
+#
+#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
+#   * `grep` fails (exit status 2)
+#   * `grep` finds `closure` (exit status 0)
+#
+# Note that `make` can in principle exit with status 1 too, but only
+# if the `-q` option is used, which should never be the case for our
+# builds, so we can assume exit status 1 only comes from `grep`.
```
jens-mueller-sociomantic: I don't see how the exit status of the first part of the pipeline could propagate to `test`.

ci/closures.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -x -o pipefail
+
+# Enable printing of all potential GC allocation sources to stdout
+export DFLAGS=-vgc
+
+# Ensure there are no lines about closure allocations in the output.
+# Note explicit check for `grep` exit status 1, i.e. no lines found.
```
jens-mueller-sociomantic: I'd suggest to shorten the comment. I find it too verbose. One could say something like:
`Passes only if make properly builds and runs and grep finds no "closure" successfully; otherwise fails.`

ci/closures.sh
joseph-wakeling-sociomantic: `set -o pipeline` is used.  So IIUC that means that if `(grep -e "closure"; test $? -eq 1)` has exit status 0, and `make` has a non-zero exit status, the latter will be the final exit status of the line.

joseph-wakeling-sociomantic: It's just a cautionary note.  In one draft I was considering putting the `test` on a follow-up line, and checking the exit status of the entire pipeline; this should in principle be valid.

jens-mueller-sociomantic: Yes. But in the comment you seem to argue that the exit code of make can somehow end up being used in the `test $? -eq 1`, don't you? My point is that the exit status of the make command cannot be used in the latter part of the pipeline. At least I don't see how it could happen for the pipeline as written. That's why I find the comment confusing. To me there are two parts of the pipeline. The latter part exits with `0` or `1` depending on the exit code returned by `grep`. The former part exits with whatever `make` exits and `pipefail` instructs the shell to return the exit code of rightmost failing part instead of just returning the exit code of the last command.

joseph-wakeling-sociomantic:
<details><summary>Updated with a simplified comment and a slightly better exit status explanation in the commit message.</summary>

```diff
diff --git a/ci/closures.sh b/ci/closures.sh
index b528930..d5b6edf 100755
--- a/ci/closures.sh
+++ b/ci/closures.sh
@@ -4,15 +4,7 @@ set -x -o pipefail
 # Enable printing of all potential GC allocation sources to stdout
 export DFLAGS=-vgc
 
-# Ensure there are no lines about closure allocations in the output.
-# Note explicit check for `grep` exit status 1, i.e. no lines found.
-# This means tests will fail if any of these conditions are met:
-#
-#   * `make` fails (exit status 2, propagated by `set -o pipefail`)
-#   * `grep` fails (exit status 2)
-#   * `grep` finds `closure` (exit status 0)
-#
-# Note that `make` can in principle exit with status 1 too, but only
-# if the `-q` option is used, which should never be the case for our
-# builds, so we can assume exit status 1 only comes from `grep`.
+# Passes only if `make` succeeds and `grep` successfully finds no
+# `closure` (exit status 1); fails if either `make` or `grep` fail
+# or if `grep` finds at least one `closure`
 beaver dlang make fasttest 2>&1 | (grep -e "closure"; test $? -eq 1)
```
</details>

ci/closures.sh
joseph-wakeling-sociomantic: As mentioned above, the `2>&1` may be necessary to handle cases where compiler output goes to `stderr` not `stdout` (this can vary by D compiler).

Personally I like the clarity of meaning of `-e "something"` ... ?

jens-mueller-sociomantic (approved):

PR #93, opened 2018-11-06
Replace all uses of Const!(...) with const(...)

joseph-wakeling-sociomantic:
Now that dmxnet is D2-only we no longer need to use the transitional template, as we can use the `const` qualifier directly.

### DIFF
--- a/integrationtest/mxnet/main.d
+++ b/integrationtest/mxnet/main.d
@@ -166,7 +166,7 @@ void main (istring[] args)
 *******************************************************************************/
 
 size_t[] predict (Symbol model, NDArray!(float) matrix_w,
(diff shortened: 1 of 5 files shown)


jens-mueller-sociomantic:
I tried counting all `Const!` to figure out whether you got all of them. But somehow it does not add up. The curious thing is that you have more replaced that I am able to find in `src/` and `integrationtest/`. What am I missing? I'm using `grep -o "Const\!"`.

joseph-wakeling-sociomantic:
The trick you can use is `git grep -ne "Const!"` ;-)

jens-mueller-sociomantic (approved):

PR #96, opened 2018-11-06
Remove or replace all imports of ocean.transition

joseph-wakeling-sociomantic:
For modules that use the `istring`, `cstring` or `mstring` aliases, we replace `ocean.transition` imports with `ocean.meta.types.Qualifiers`.  In all other modules the `ocean.transition` import is just dropped, as dmxnet no longer uses any of the functionality it defines.

### DIFF
--- a/integrationtest/mxnet/MNIST.d
+++ b/integrationtest/mxnet/MNIST.d
@@ -32,8 +32,8 @@ import ocean.io.compress.ZlibStream;
 import ocean.io.device.File;
 import ocean.io.Path;
 import ocean.math.Math;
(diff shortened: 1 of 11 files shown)


christoph-wasicki-sociomantic (approved):

jens-mueller-sociomantic:
An alternative here could have been to replace `istring` with `string`, `cstring` with `const(char)[]` and `mstring` with `char[]`. Then the import could be removed entirely.

joseph-wakeling-sociomantic:
Yes, but that's non-idiomatic within our library space.  We can discuss that for the future but for now at least probably best to go with existing standards.

PR #97, opened 2018-11-06
Encapsulate imports required for debugReport helper function

joseph-wakeling-sociomantic:
Imports from `core.sys.posix.unistd` and `ocean.text.convert.Formatter` have been encapsulated inside the `debugReport` function, since they are unused by any other part of the `mxnet.Handle` module.  This is a little bit bikesheddy, but ensures that debug helper functionality is cleanly separated from production code.

The import of `ocean.io.Stdout` is no longer used (it's just a hangover from an earlier implementation), so we just drop it.

### DIFF
--- a/src/mxnet/Handle.d
+++ b/src/mxnet/Handle.d
@@ -15,11 +15,7 @@ module mxnet.Handle;
 
 import mxnet.Exception;
 
-import core.sys.posix.unistd;
-
-debug (MXNetHandle) import ocean.io.Stdout;
 import ocean.meta.types.Qualifiers;
-import ocean.text.convert.Formatter;
 
 version (unittest)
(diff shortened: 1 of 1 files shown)


joseph-wakeling-sociomantic:
Tagging this as D2 simply because it's a nice bit of separation-of-concerns that is only possible thanks to D2 nested imports.

jens-mueller-sociomantic (approved):
One could separate the removing of `ocean.io.Stdout` into a different patch. Because it seems to be primarily cleanup whereas the latter looks more like refactoring.

joseph-wakeling-sociomantic:
Yes, but I think in context that's a bit too fussy.  Let's take this in.

PR #98, opened 2018-11-07
Add a .keep file for the relnotes folder

joseph-wakeling-sociomantic:
This will ensure that the `relnotes` folder is preserved by neptune when clearing its contents after a release.


alaksiej-stankievicx-dunnhumby (dismissed):

jens-mueller-sociomantic:
Is this something that neptune does or it is more like that this way you ensure that there will be always a file inside `relnotes`, i.e., git will keep tracking the directory.

joseph-wakeling-sociomantic:
The latter.  It means git will still keep track of the folder when neptune clears out its other contents.  I've updated the commit message accordingly.

alaksiej-stankievicx-dunnhumby (approved):

jens-mueller-sociomantic (approved):

PR #100, opened 2018-11-07
Use enum for manifest constants

jens-mueller-sociomantic:
This change replaces all uses of `static immutable` with `enum` where
the use of manifest constant won't cause an allocation. Specifically,
this is the case when defining manifest constants using non-array
literals (which includes string literals).

When converting to D2-only (see 6b81096) D1 `const` was replaced by
`static immutable`. In all these cases D1 `const` served as a manifest
constant. In D2 `enum` is used to define manifest constants. This is
preferred except when done for array literals accessed more than once
when running. Because in this case each use of the manifest constant will
cause an allocation which is often not wanted. This does not apply to
string literals which are treated specially. Hence, using an `enum` for
a string literal carries no risk of allocation when used and is usually
fine. Using `enum` with this in mind enables replacing `static immutable`
which is clearer in intent and avoids any possible need for static memory
to hold the immutable data.

### DIFF
--- a/src/mxnet/Util.d
+++ b/src/mxnet/Util.d
@@ -49,14 +49,14 @@ version (unittest)
 
 public mstring toNoLossString (in float x, mstring buf)
 {
-    static immutable max_length = 1 + // optional minus sign
-                       1 + // leading decimal digit
-                       1 + // .
-                       8 + // 8 decimal digits
-                       1 + // e
-                       1 + // sign
-                       2;  // decimal exponent
-    static immutable format_string = "{:e8}";
+    enum max_length = 1 + // optional minus sign
+                      1 + // leading decimal digit
+                      1 + // .
+                      8 + // 8 decimal digits
+                      1 + // e
+                      1 + // sign
+                      2;  // decimal exponent
+    enum format_string = "{:e8}";
 
     enforce(buf.length >= max_length);
     return snformat(buf, format_string, x);
--- a/src/mxnet/NDArray.d
+++ b/src/mxnet/NDArray.d
@@ -822,7 +822,7 @@ public alias Tuple!(float, double, int, ubyte) SupportedElementTypes;
 
 public template isSupportedElementType (T)
 {
-    static immutable bool isSupportedElementType =
+    enum bool isSupportedElementType =
         IndexOf!(T, SupportedElementTypes) < SupportedElementTypes.length;
 }
 


src/mxnet/Symbol.d
```diff
@@ -575,7 +575,7 @@ public class Activation : Symbol
     }
     body
     {
-        static immutable istring[] activations = ["relu", "sigmoid", "softrelu", "tanh"];
+        enum istring[] activations = ["relu", "sigmoid", "softrelu", "tanh"];
```
mihails-strasuns-sociomantic: This will cause new array  to be allocated each  time constant is used

jens-mueller-sociomantic: What does used mean here? In the later code it is used only once. But I think your point is that depending on the how it is used it may be more than once. I check the other uses of `enum` as well.

jens-mueller-sociomantic: This was the only one.

jens-mueller-sociomantic: I fix it.

jens-mueller-sociomantic: I think I'm also going to clarify the documentation. The important thing is how often the array may be referenced at run-time.

src/mxnet/Util.d
```diff
@@ -49,14 +49,14 @@ version (unittest)
...
-    static immutable max_length = 1 + // optional minus sign
+    enum max_length = 1 + // optional minus sign
                        1 + // leading decimal digit
                        1 + // .
                        8 + // 8 decimal digits
                        1 + // e
                        1 + // sign
                        2;  // decimal exponent
-    static immutable format_string = "{:e8}";
+    enum format_string = "{:e8}";
```
joseph-wakeling-sociomantic, line 59: Is this not also going to create the same problem of an allocation per use?

jens-mueller-sociomantic: I think not (see https://wiki.dlang.org/Declaring_constants).

jens-mueller-sociomantic: So it does not apply to `string`s. But the commit message does not cover it yet.

nemanja-boric-sociomantic: Yeah this is not going to create a problem, because string literals are not created using GC.

jens-mueller-sociomantic: I wonder why the compiler cannot create some read-only memory for any array defined by `enum`. Why does it only work for `string`s?

nemanja-boric-sociomantic: Well the string literals are readonly, but `[1, 2]` is not :-)

jens-mueller-sociomantic: True. But the compiler also sees the `enum` so it could infer that there is no way to change such an compile-time array.

jens-mueller-sociomantic:
I'm going to improve the commit message. Any more feedback?

jens-mueller-sociomantic:
Updated commit message and rebased.

joseph-wakeling-sociomantic:
Your commit message refers to `non-array literals`, can you note the string literal case as well, since technically a string could be interpreted as an array?

src/mxnet/Util.d
```diff
@@ -49,14 +49,14 @@ version (unittest)
 
 public mstring toNoLossString (in float x, mstring buf)
 {
-    static immutable max_length = 1 + // optional minus sign
+    enum max_length = 1 + // optional minus sign
```
joseph-wakeling-sociomantic, line 52: Could you correct all the indentation here in a follow-up patch?

jens-mueller-sociomantic:
> Your commit message refers to `non-array literals`, can you note the string literal case as well, since technically a string could be interpreted as an array?

You mean making clearer that string literals are non-array literals? I used the term `literal` for that reason. Because `string`s are arrays but string literals are not array literals. But I see the potential for confusion.

jens-mueller-sociomantic:
Rephrase commit message and fixed indentation.

src/mxnet/Symbol.d
mihails-strasuns-sociomantic: > What does used mean here?

Every time array literal expression is hit in the code during runtime execution, new array gets allocated. When you use `enum`, constant name is replaced with array literal (as in copy-paste) so the same rule applies.

For example:

```D
enum arr = [ 1, 2, 3 ];
foreach (i; 0..1)
{
    writeln(arr); // will allocate new array 10 times total
}
```

src/mxnet/Util.d
mihails-strasuns-sociomantic: That is the trick - you can:

```D
    auto x = [ 1, 2, 3 ];
    x[0] = 42;
```

For this to work, array literals must allocate mutable memory. String literals are immutable thus they can be optimized to static read-only memory.

jens-mueller-sociomantic: Yes, the compiler does not look at it from a higher level (stops at the literal). And it seems to be implemented as a special case for `string`s. Because it also does not seem to work for `[cast(immutable(int)) 1]`. Maybe just implementation, maybe a principle problem.

mihails-strasuns-sociomantic: Yeah, it is implementation defined. For strings, specification requires it - for anything else it is an optional valid optimization.

joseph-wakeling-sociomantic (approved):
Given feedback from everyone I'm happy to take this in.  Thanks folks!

joseph-wakeling-sociomantic (approved):

PR #106, opened 2019-04-03
[WIP] Update C API to mxnet v1.4.0

daniel-zullo:


src/mxnet/c/c_api.d
```diff
@@ -51,36 +53,51 @@ extern (C):
 /*! \brief MXNET_DLL prefix for windows */
 
 /*! \brief manually define unsigned int */
-alias uint mx_uint;
+alias mx_uint = uint;
```
alaksiej-stankievicx-dunnhumby, line 56: I think change `alias` for more D2 syntax should be done in separate commit.

src/mxnet/c/c_api.d
```diff
@@ -104,7 +121,13 @@ struct NDArrayOpInfo
     bool function (int, int*, uint**, void*) infer_shape;
     bool function (char***, void*) list_outputs;
     bool function (char***, void*) list_arguments;
-    bool function (const(int)*, const(int)*, const(int)*, int*, int**, void*) declare_backward_dependency;
+    bool function (
```
alaksiej-stankievicx-dunnhumby, line 124: and this is also stylistic changes should be moved to separate commit.

alaksiej-stankievicx-dunnhumby (commented):
You should split style update vs real API update, because now it is difficult to track it.

src/mxnet/c/c_api.d
daniel-zullo: yes, that's right. Jens explain me the way to update it and we will address previous versions first

src/mxnet/c/c_api.d
daniel-zullo: I have found the right `dstep` command arguments to avoid these changes

daniel-zullo:
Closing as previous versions will be addressed first.

PR #108, opened 2019-04-05
Add reporting of MXNet version

jens-mueller-sociomantic:
This change provides a function to obtain the MXNet version of the
linked library. We also output this version as a unittest. The
diagnostic can be useful when checking which MXNet version is actually
used.


src/mxnet/API.d
```diff
@@ -157,3 +159,150 @@ public void notifyShutdown ()
...
+        Patch version
+
+    ***************************************************************************/
+
+    ubyte patch;
+
+
+    /***************************************************************************
+
+        Formats this version
```
daniel-zullo: You might want to change it to `Formats the MXNet version`

src/mxnet/API.d
```diff
@@ -157,3 +159,150 @@ public void notifyShutdown ()
...
+
+public struct MXNetVersion
+{
+    /***************************************************************************
+
+        Major version
+
+    ***************************************************************************/
+
+    uint major;
```
daniel-zullo, line 194: would it be possible to use `///` to document member variables or single documentation lines?

daniel-zullo (commented):
I only have a cosmetic suggestion and a question and LGTM

src/mxnet/API.d
```diff
@@ -157,3 +159,150 @@ public void notifyShutdown ()
...
+        major = specifies the major version
+        minor = specifies the minor version
+        patch = specifies the patch version
+
+    Returns:
+        the specified MXNet version
+
+*******************************************************************************/
+
+private MXNetVersion mxnetVersion (uint major, ubyte minor, ubyte patch)
```
stefan-koch-sociomantic: maybe you could get rid of this helper?

jens-mueller-sociomantic: Why? I'm using it in the tests.

src/mxnet/API.d
jens-mueller-sociomantic: I could.

src/mxnet/API.d
stefan-koch-sociomantic: it does not look like you really need it.
You can do the modding and division in struct literal as well, no?

jens-mueller-sociomantic: I don't see it.

jens-mueller-sociomantic: Clarified in real life. The suggestion is to use the implicit constructor.

src/mxnet/API.d
```diff
@@ -157,3 +159,108 @@ public void notifyShutdown ()
...
+{
+    // major * 10_000 + minor * 100 + patch
+    // see `MXNET_VERSION` is defined in `include/mxnet/base.h`
+    MXNetVersion version_ =
+    {
+        major : mxnet_version / 10_000,
+        minor : mxnet_version % 10_000 / 100,
+        patch : mxnet_version % 100,
+    };
+    return version_;
```
jens-mueller-sociomantic, line 244: I could also do

```d
return MXNetVersion(mxnet_version / 10_000,
                    mxnet_version % 10_000 / 100,
                    mxnet_version % 100);
```

but with the `{ ... }` it looks more explicit though easier to break when we add a field.

daniel-zullo: The current version is easier to read and it is unlikely a new field will be added to the version

daniel-zullo (dismissed):
LGTM

src/mxnet/API.d
jens-mueller-sociomantic: I'll agree.

jens-mueller-sociomantic:
Add missing release notes.

UplinkCoder (approved):
LGTM

daniel-zullo (approved):

PR #111, opened 2019-04-08
Fix reporting of testing accuracy in percentage

jens-mueller-sociomantic:
The accuracy for the MNIST integration test reports in percentages of
correctly classified hand-written digits. But the calculation is missing
a multiplication by one hundred to be proper which is fixed in the
change.


UplinkCoder (dismissed):
LGTM

daniel-zullo (approved):
LGTM

PR #113, opened 2019-04-15
Remove unnecessary conversion to array

jens-mueller-sociomantic:
This change removes the unnecessary conversion of the outputted
predicted probabilities for the integration test. This conversion is not
needed because we can directly refer to the returned data.


alaksiej-stankievicx-dunnhumby (approved):
LGTM

daniel-zullo (approved):

PR #115, opened 2019-09-11
[WIP] Add Ubuntu bionic support

daniel-zullo:
Only Ubuntu xenial was supported.


daniel-zullo:
First needs to update and generate a package in https://github.com/sociomantic-tsunami/mxnet

